PR #22, opened 2014-09-08
Use _cpt AND cpt ...

mohierf:
... take care of ascending compatibility.
Added a test for 'cpt' variable because of ascending compatibility.

Moreover, I added the same tests when receiving data for WMI and files collect ...

Added extra-debug logs for registry content when updating computer data.

Added extra-debug logs for files collection and fixed a bug : only one file was stored with computer !

Added regexp assignement feature from registry values for Status and Location fields.


ddurieux:
Yet fixed in other commit

PR #27, opened 2015-05-10
Fix right for link with plugin monitoring

ddurieux:


kiniou:
I was not aware of that glpi-monitoring code in glpi-fusioninventory parts and i'm really not fond of it. This complexify the maintenance of fusioninventory and made it tightly dependant of glpi-monitoring.

We could use some defined hooks that are registered from external plugins and automatically detected by fusioninventory plugin setup.

Meanwhile, i will merge this PR since it fix compatibility between those plugins.

PR #37, opened 2015-06-02
Enhancement to support CRSF compliant POST method for collect tasks

g-bougard:
This enhancement is needed to enable POST method for 2.3.x agents collect task enhancement proposed in https://github.com/fusioninventory/fusioninventory-agent/pull/27 pull request

### DIFF
--- a/b/collect/index.php
+++ b/b/collect/index.php
@@ -67,6 +67,10 @@ if (isset($_GET['action'])) {
                            $out = $class->run($data, $a_agent);
                            if (count($out) > 0) {
                               $response['jobs'][] = $out;
+                              // Inform agent we request POST method, agent will then submit result
+                              // in POST request if it supports the method or it will continue with GET
+                              $response['postmethod'] = 'POST';
+                              $response['token'] = Session::getNewCSRFToken();
                            }
                            $pfTaskjobstate->changeStatus(
                                    $data['id'],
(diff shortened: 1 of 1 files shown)


ddurieux:
Why need POST method? GET works too no?

g-bougard:
POST method in not a need, just and enhancement here. I'm also server admin and don't like how GET method can fulfill HTTP server access log. The POST method can be used to limit this minor issue.  Right now, this enhancement don't limit the number of posted values, just their size in server logs.

I think this can also be a security issue against collected values if access log can be read by not authorized user. The POST method fix this too.

kiniou:
Hi @g-bougard,

I think you are trying to solve some non-issue here. The choice of GET method for agent requests has been specified to this purpose of logging each requests.

If your concerns are filtering out some sensitive logs from your webserver's logs, there are already mechanisms to do it from the webserver level (like [CustomLog](http://httpd.apache.org/docs/2.2/mod/mod_log_config.html#customlog) for apache or [this gist](https://gist.github.com/deltheil/2867074) for nginx).

However, if you can come up with some real security test case (as i can't find any at the moment myself), we can reconsider this enhancement.

g-bougard:
Hi kiniou,

you're right, this patch is really not fixing any issue.

To be accurate, this patch only enables POST method for collect tasks and only if agent support it. The PR is also to be associated with the related fusioninventory-agent PR as reference.

In fact, requests number is even not changed : it only permit the agent to specify a postmethod=POST parameter on each request so collected valued can be read from the POST content. IMHO, this drastically reduces log entropy in the case of heavy collecting tasks.

For my information, can you point me to the specification which report the choice of GET for the purpose of logging each requests ? I would imagine such need may also been implemented in a separated journal.

I agree filtering out sensitive information changing http server configuration is a way of doing. Btw I really think in the default case collected values shouldn't appear in http server.

The real enhancement would be to add a configuration parameter to set the way to post collected values.

About security test case, what kind of test would you expect ? My security concern was only about sensitive values logged in a place not designed for it.

ddurieux:
You can see here: http://www.fusioninventory.org/documentation/dev/spec/protocol/collect/

ddurieux:
ping ?

ddurieux:
Merged manually

g-bougard:
To keep the ref, I see it was merged in 7378c17f21081fe41a6e1a397a3ec44d1a3e6dab
Thank you

PR #39, opened 2015-06-03
fix task display with networkinventory

orthagh:
Tasks with networkinventory jobs now displays correctly the targets with an associated range


orthagh:
Edit 2015/0605 : 
 - setup.php optimisation (load js/css only when needed)
 - js libs upgrade (small rendering gain)
 - divide by 3 dom rendering on task (389 equipments : 50s to 12-15s)

ddurieux:
Many problem I have seen when I have tested a little:

* Some tasks are empty in general monitoring log page, but have something in master branch
* timeslot select in task form not appear

So perhaps other things, but not have more tested because I have these 2 problems

orthagh:
Indeed, there was a missing commit.

ddurieux:
Not works enough, I have only the sucessfull data, not the prepared, running...in the monitoring log tab of inventory task

Efe16:
Hello, I did some test whose results are on the French Fusion Forum : http://forum.fusioninventory.org/viewtopic.php?pid=18345#p18345

orthagh:
@Efe16 can you test with my full bugfixes branch  : https://github.com/orthagh/fusioninventory-for-glpi/tree/0.85/teclib

Efe16:
Hello,

I've test this version https://github.com/orthagh/fusioninventory-for-glpi/tree/0.85/teclib and I have the same result :(
I've test old version  of client 
- V2.3.15 : same bug of V2.3.16 but the client stop to the scan IP 
- V2.3.12 : the client crash
- V2.2.7.3 : the client is not compatible with the server for the scan IP.

Another small bug with this version : 
In Job Executions, if I put the refresh to 5 second, the animation of the refresh is OK but no the result of the page ; if I refresh the page, the result is OK.  

I Put the result in french to  http://forum.fusioninventory.org/viewtopic.php?pid=18345#p18345

ddurieux:
Merged manually

PR #44, opened 2015-06-19
0.85/collapse tasks in monitoring / logs

orthagh:
proposal for http://forge.fusioninventory.org/issues/2901


ddurieux:
Feature is a good idea, but when refresh all 5 secondes, the task expanded before is collapsed after each refresh

orthagh:
ok good point, i'll check this this week

orthagh:
Here (and rebased)

ddurieux:
related #1632

PR #1702, opened 2015-10-21
fix warnings in computer list of dynamic group

orthagh:
And add new methods to improve performance


ddurieux:
Rebased manually to commit manually

PR #1720, opened 2015-11-16
stats on board is now with entities

orthagh:
Conflicts:
	inc/menu.class.php


ddurieux:
Replaced by PR #1762 because queries not better optimized

PR #1730, opened 2015-12-08
retrieve inventory data in case of delete last lock on a item

orthagh:
This impacts the retrieval of inventory data when deleting locks.

If after deletion of a lock, another locks are still present, it works as intended (update case)
But in case of deletion of all lock (or last lock), deletion is ok but retrieval of data not triggered (delete case).

This fix, check the presence of item->update in deletelock function to detect if a CommonDBTM::delete or update function is triggered.

### DIFF
--- a/inc/lock.class.php
+++ b/inc/lock.class.php
@@ -749,6 +749,12 @@ class PluginFusioninventoryLock extends CommonDBTM{
             }
          }
       }
+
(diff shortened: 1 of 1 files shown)


ddurieux:
Can you rebase with master please? I have made modifications with travis (unit tests).
``` git rebase master ```

PR #1754, opened 2016-01-22
Rework FI tabs display in assets

wawax:
Try to minimize the number of tabs added by FusionInventory for an asset : 

- remove tab for agent status & wakeup. Replace it by a tootip on the asset's main page
- remove remotemanagement tab : display it on the asset's main page 
- do not show log or rules matched when computer is not managed by fusioninventory
- do not show task logs when computer is not managed by fusioninventory


ddurieux:
Not sure it's all ok, because with your patch, you can't be able to lock a field on a computer created manually and before install agent on computer

trasher:
What about adding a parameter in the plugin global configuration for that?
That would permit the user to display back not managed devices if he wants to.

ddurieux:
It could be a good idea to make specs (enumerate tabs and define when display it or not) ;)

trasher:
Hum... OK. I'll take a look; I'm not yet enough familiar with the project to estimate this :)

wawax:
@ddurieux I was also thinking at merging the 3 collect tabs for a computer into one.
Do you think that it's a good idea ?

wawax:
@ddurieux is PluginFusioninventoryInventoryComputerStorage still in use ? It seems not to be used anymore.
Should we remove this class ?

ddurieux:
There was code for that, but 3 years ago Goneri say me disable all the code.
In fact not sure we have yet it in the repository...

ddurieux:
For the collect, it's displayed only when have data. The main problem is than you can have different information in each and can have very many information, so refer keep them in 3 distinct tabs ;)

wawax:
@ddurieux you're right for the collect.
I'll drop the code for computerstorage, as it is unused.

wawax:
I've splitted the PR in sevaral smaller PRs.

see https://github.com/fusioninventory/fusioninventory-for-glpi/pull/2230 , https://github.com/fusioninventory/fusioninventory-for-glpi/pull/2228 , https://github.com/fusioninventory/fusioninventory-for-glpi/pull/2231

PR #1755, opened 2016-01-22
Delete locks and matches rule logs when an asset is purged

wawax:
Clean locks and rules when an asset is purged


ddurieux:
Need a test for that

PR #1795, opened 2016-04-18
Add hooks to import data from plugins

wawax:
Add hooks for plugins to get inventory data from the XML file

### DIFF
--- a/inc/formatconvert.class.php
+++ b/inc/formatconvert.class.php
@@ -293,6 +293,13 @@ class PluginFusioninventoryFormatconvert {
          $a_inventory['fusioninventorycomputer']['operatingsystem_installationdate'] = "NULL";
       }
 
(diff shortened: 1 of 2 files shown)


inc/formatconvert.class.php
```diff
@@ -293,6 +293,13 @@ static function computerInventoryTransformation($array) {
          $a_inventory['fusioninventorycomputer']['operatingsystem_installationdate'] = "NULL";
       }
 
+      $plugin_values = Plugin::doHook("fusioninventory_addinventoryinfos",
```
ddurieux: Perhaps better to add the hook after the _$a_inventory_ is completely filled?

wawax:
A new one has been created : https://github.com/fusioninventory/fusioninventory-for-glpi/pull/1871

PR #1817, opened 2016-05-20
Add memory support when there is no physical memory detected, e.g. w…

### LINKED ISSUE #1020: HARDWARE/MEMORY ignored if now MEMORIES
---

Author Name: **Gonéri Le Bouder** (Gonéri Le Bouder)
Original Redmine Issue: 1937, http://forge.fusioninventory.org/issues/1937
Original Date: 2013-02-13

---

If there is no <MEMORIES /> entry, we should create a “dummy memory slot” with the HARDWARE/MEMORY. without that, it's not possible to get the information directly from GLPI.

guillaumelecerf:
…ith a virtual machine

Fixes #1020

### DIFF
--- a/inc/formatconvert.class.php
+++ b/inc/formatconvert.class.php
@@ -846,6 +846,13 @@ class PluginFusioninventoryFormatconvert {
                }
             }
          }
+         else if (isset($array['HARDWARE']['MEMORY'])) {
+            $array_tmp = $thisc->addValues($array['HARDWARE'],
+                                           array(
+                                              'MEMORY' => 'size'));
+            $array_tmp['designation'] = 'Dummy Memory Module';
+            $a_inventory['memory'][] = $array_tmp;
+         }
       }
 
       // * MONITORS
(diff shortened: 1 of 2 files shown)


ddurieux:
Seems a good idea, but unit tests not pass, so need check and update them if required

guillaumelecerf:
AFAIK unittests are already broken, so trying to determine if my patch broke them more is rather difficult.

ddurieux:
yes I have a problem with the version 5.4 of PHP and glpi 0.90, but other may be ok.
You can check it: https://travis-ci.org/fusioninventory/fusioninventory-for-glpi/jobs/131636819

Of course, I can help you for that ;)

Seems a problem in _ComputerTransformation::ComputerGeneral_ seems logical because with your patch you can add a momory entry in GLPI in the case you haven't <MEMORIES> section but have memory in <HARDWARE> section. So in this case, modify the unit test.

guillaumelecerf:
Unittests should be OK now, thanks for the hint ;)

ddurieux:
It's not ok because I have passer master branch for next version of GLPI (in dev) but your code seems ok. I will merge it tomorrow in master + bugfix/0.90 ;)

ddurieux:
Thanks ;)

PR #1819, opened 2016-05-22
Fix a cycle in the importToken method

poum:
I experienced memory leak calling from sccm plugin and it seems to me that this line was creating an useless cycle .... right ?


ddurieux:
What the link with sccm? not understand

trasher:
I agree:: object should not be instanciated twice, this is not needed. But the proposed fix is wrong, and out of date. See #2077

PR #1839, opened 2016-05-29
Disable xdebug during tests

guillaumelecerf:


ddurieux:
Not work, we have in travis ```The Xdebug extension is not loaded. No code coverage will be generated.```

guillaumelecerf:
What a shame, because the speedup is huge.
Closing.

ddurieux:
Yes but in case errors, not sure have all the trace

PR #1874, opened 2016-07-12
 Delete locks and matches rule logs when an asset is purged

trasher:
Clean locks and rules when an asset is purged.

Replaces #1755, to add requested unit tests.

### DIFF
--- a/phpunit/2_Integration/ComputerLicenseTest.php
+++ b/phpunit/2_Integration/ComputerLicenseTest.php
@@ -198,7 +198,7 @@ class ComputerLicenseTest extends RestoreDatabase_TestCase {
 
       //Second, clean and check if it has been removed
       $pfComputerLicenseInfo = new PluginFusioninventoryComputerLicenseInfo();
-      $pfComputerLicenseInfo->cleancomputer(1);
+      $pfComputerLicenseInfo->cleanComputer(1);
 
       $pfComputerLicenseInfo->getFromDB(1);
       $this->assertEquals(0, count($pfComputerLicenseInfo->fields));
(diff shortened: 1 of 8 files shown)


ddurieux:
Can you rebase please ?

trasher:
I've just rebased :) 
Sorry for the delay, I had a motorbike accident and I've been away from a computer for a while...

ddurieux:
Thanks ^^ will read it in few hoors
No problem for the delay ;)

phpunit/2_Integration/ComputerLicenseTest.php
```diff
@@ -160,8 +160,48 @@ public function Licenses() {
...
+      $this->assertEquals($a_ref,
+                          $pfComputerLicenseInfo->fields,
+                          'License data');
+
+      //Second, clean and check if it has been removed
+      $pfComputerLicenseInfo = new PluginFusioninventoryComputerLicenseInfo();
+      $pfComputerLicenseInfo->cleancomputer(1);
+
+      $pfComputerLicenseInfo->getFromDB(1);
+      $this->assertEquals(0, count($pfComputerLicenseInfo->fields));
```
ddurieux: Not total agree with tests :p, you test the function cleancomputer() in the class, but need too in test delete/purge the computer and check if data are deleted in collects and licenses

trasher:
If it sounds good to you for Collects, I'll do the same on Licences.

ddurieux:
Yes it's nice like it ;)

ddurieux:
Miss documentation for some functions and other not in right format, see an example here: https://github.com/fusioninventory/fusioninventory-for-glpi/blob/master/inc/collect.class.php#L77-L83

otherwise, it's ok

trasher:
OK, I'll fix that, thank you!

Issue #1889, opened 2016-07-26
Many problems in collect 


ddurieux:
In collect, have many problems:
* registry data not linked to registry id
* unable to collect more than 1 registry information (same for wmi and files)
* ...

PR #1890, opened 2016-07-26
[wip] Begin fix collect part. references #1889

ddurieux:


ddurieux:
todo:

* add unit tests
* add code to be compatible with old version of agent collect module

PR #1913, opened 2016-08-29
Add a build parameter to replace hardcoded http host in CollectsTest.php

trasher:
There is an hardcoded http host and port hard coded in CollectsTest.php file; that prevents unit tests to be run on my local machine.

I've tried to add a parameter at runtime; I do not know if there is another way to go, I do not know much ant nor phpunit, and I'm not sure of the possible interactions between both of them.

Well, that works for me, but maybe there would be a more simple or elegant  solution.

### DIFF
--- a/phpunit/build.xml
+++ b/phpunit/build.xml
@@ -6,6 +6,7 @@
  >
 
  <property name="fusioninventory.directory" value="${basedir}/glpi/plugins/fusioninventory" />
+ <property name="fusioninventory.http.host" value="localhost:8088" />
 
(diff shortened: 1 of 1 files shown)


ddurieux:
unit tests not pass :/

trasher:
Yep, I've seen; but I've not yet understood why :/

trasher:
OK, unit tests have been fixed: I do not understand why this fail on travis and works locally; the solution was to set ant's move not to fail on error...

PR #1915, opened 2016-08-30
Use core antivirus stuff

trasher:
Since antivirus stuff do exists in GLPI core, we no longer need specific code in the plugin for that now.


ddurieux:
1/ don't rename fields on migration, because you will drop after, so it will use times for that and not performant

2/ we need add some tests about antivirus add / update from inventory received

trasher:
You're right, I'll drop the rename :)

And, OK, I'll work on some unit tests.

trasher:
I've drop the rename of the fields, and I've added some unit tests. Lemme know if it sounds good to you ;)

ddurieux:
yep it's ok ;)

Thanks

trasher:
Great, thank you :)

PR #1919, opened 2016-09-02
Restart jobs feature

trasher:
Add possibility to restart jobs


trasher:
OK, we did not understand that jobs were restated via crontabs even if they're in error or cancelled; so this PR is useless :)

PR #1922, opened 2016-09-06
Use core's architectures tables

trasher:
Use core's architectures table instead of specific fusion one.
Also remove specific table; and take care of updating data.


trasher:
Closed for now, waiting for #1923 to be merged, since some changes are needed on the same files/lines

PR #1923, opened 2016-09-07
Use core's tab to display fusioninventory OS informations

trasher:


trasher:
Related to glpi-project/glpi#919

hook.php
```diff
@@ -2315,4 +2315,135 @@ function plugin_fusioninventory_getDatabaseRelations() {
...
+                   echo "<tr class='tab_bg_1'>";
+                   echo "<td>".__('Operating system')."</td>";
+                   echo "<td>";
+
+                   $os = new OperatingSystem();
+                   if ($os->getFromDB($pfComputerOperatingSystem->fields['operatingsystems_id'])) {
+                      echo $os->fields['name'];
+                   } else {
+                      echo '-';
+                   }
```
ddurieux: Perhaps replace these lines by Dropdown::getDropdownName(), it will simplify code and reduce it.
So same for other fields in this file

trasher:
Ooooh, OK, did not know about this one, thanks; i will do :)

Issue #1529, opened 2015-09-30
Put in trash software with no install and and license


ddurieux:
---

Author Name: **Walid Nouh** (@wawax)
Original Redmine Issue: 2743, http://forge.fusioninventory.org/issues/2743
Original Date: 2014-10-07

---

It may be interesting to push in trash a software without install & license.
It is the default behavior of OCS Sync for years.

Advantages :
- list only software really in use
- it's easy to clean old softwares

Maybe add an option to set or not this behavior ?

ddurieux:
this version is no longer supported

PR #1945, opened 2016-10-20
1529 trash soft

trasher:
Fix for ticket #1529 

This PR contains commits from #1931 to avoid conflicts...


wawax:
Hi @trasher ,

This works but if the software, which is in the trash appears again in the inventory, it doesn't seems to restore it from the trash.

Thanks,

ddurieux:
it's in waiting because this patch can have strange behavior in multiple inventories in same time, so I must have a big thing about this

PR #1950, opened 2016-10-30
Fix footer because too intruisive. closes  #1938

### LINKED ISSUE #1938: Intrusive footer
Since commit 66e03f11, an additional footer is added - visually - above the GLPi one.

There are several issues with that:
- it is added after the page ends (ie. after the `</html>` tag because `register_shutdown_function` has been used), and therefore, this is not html standards compliant at all,
- it creates a second `<div id="footer">` while html ids must be unique,
- it is on some cases added on some core or plugins pages it should not, that breaks (at least) display
- this does not fit changes made on the core's footer.

I do not get the point for adding this, the commit message lacks of information. Anyways, what if all plugins authors decide to do the same based on this example?

I was thinking about a way to fix the display issues, but I wonder if this should not be just removed...

ddurieux:


js/footer.js
```diff
@@ -0,0 +1,16 @@
+var fifooter = "<div id='footer'> \
```
trasher: There still is a duplicate ID, they must be unique.

Why not just add the `a` content in the last cell of the `#footer` with a `<br/>`?

ddurieux:
Because there is a CSS on the id _footer_

trasher:
Take a look with:
```
diff --git a/js/footer.js b/js/footer.js
index b7fe606..ff4d356 100644
--- a/js/footer.js
+++ b/js/footer.js
@@ -1,16 +1,12 @@
-var fifooter = "<div id='footer'> \
-<table width='100%'>  \
-<tr> \
-<td class='right'> \
+var fifooter = "<br/> \
 <a class='copyright' href='http://fusioninventory.org/'> \
 FusionInventory 9.1+1.0 - Copyleft \
 <span style='display:inline-block;transform: rotate(180deg);font-size: 12px;'>&copy;</span> \
 2010-2016 by FusionInventory Team \
-</a> \
-</td> \
-</tr> \
-</table> \
-</div>"
+</a>";
 
-$(window).bind("load", function() { $("#footer").after(fifooter); })
+$(window).bind("load", function() {
+   $('#footer').css('height', 'auto');
+   $("#footer td.right").append(fifooter);
+});
```

Does that look ok for you? Do not seem the existing CSS cause any issue.

ddurieux:
I've fixed with your recommandation ;)

trasher:
Unit tests are failing, but this is not due to footer changes; so: approved!

PR #1958, opened 2016-11-08
Fix call to addinventoryinfos: using doHook to not return correctly t…

wawax:
Calling doHook in formatconvert do not retunr correctly the plugin's values to later process.
It's mandatory to call doHookFunction, which only return plugin's values to be added.


ddurieux:
Thanks for the test ;)

PR #1997, opened 2016-12-08
Fix search (dynamic and static) in fusioninventory groups

ddurieux:


front/deploygroup.form.php
```diff
@@ -50,22 +50,23 @@
 
 $group = new PluginFusioninventoryDeployGroup();
 
-if (isset($_POST['save'])) {
+if (isset($_GET['save'])) {
+   Toolbox::logdebug($_GET);
```
trasher: Really? ;)

PR #2007, opened 2016-12-19
Connecting IP Phone to PC, Device, etc.

jsamaniegog:


inc/formatconvert.class.php
```diff
@@ -2050,7 +2050,9 @@ static function networkequipmentInventoryTransformation($array) {
                                                     'SYSDESCR' => 'sysdescr',
                                                     'MODEL'    => 'model',
                                                     'IP'       => 'ip',
-                                                    'SYSMAC'   => 'mac',
+                                                    //'SYSMAC'   => 'mac',
+                                                    'SYSMAC'   => 'sysmac',
+                                                    'MAC'      => 'mac',
```
ddurieux: Not agree with that, if we are in `CDP`, we don't have the `MAC` field in the XML, see http://fusioninventory.org/documentation/dev/spec/protocol/netinventory.html

ddurieux (changes requested):

inc/formatconvert.class.php
jsamaniegog: That's the problem, when we are in CDP we don't know the mac addresses connected to the port, for this reason i made a pull request in fusioninventory-agent, check this: https://github.com/fusioninventory/fusioninventory-agent/pull/201/files

trasher:
Is this old PR still actual? Referenced agent PR has not been integrated.

g-bougard:
A fix included in agent 2.4.1 finally sets SYSMAC in place of MAC for SIP phone while parsing CDP infos (see closed linked agent PR). @jsamaniegog should test it and tell if this answer is still needed or if his trick does a better job.

PR #2041, opened 2017-01-30
Required changes to be compatible with GLPI 9.2

trasher:


phpunit/RestoreDatabase_TestCase.php
```diff
@@ -1,8 +1,8 @@
-<?php
+n<?php
```
orthagh: !

trasher: Erf...

trasher:
As for the frequencies, yes, this is mandatory. A PR have been merge on the 9.1 branch to drop this extra data in the FormatConvert class, but unit tests are populating directly data.
Also, with the " MHz", we got a warning with PHP 7.1 from the core.

orthagh:
ok fine for me :)

PR #2044, opened 2017-02-01
Link agent into ignoredimportdevice

orthagh:
A description:
- add a new key into glpi_plugin_fusioninventory_ignoredimportdevices table for agents_id
- add a new searchoption corresponding to this key
- update existing ignored import device row when a similar entry exist (instead of always adding it)
- minor: [remove a duplicate in a searchoption of this class](https://github.com/fusioninventory/fusioninventory-for-glpi/pull/2044/files#diff-c745a6385b9b7ba9fc4e5e79f0e78c41L116)

### DIFF
--- a/inc/inventorycomputerinventory.class.php
+++ b/inc/inventorycomputerinventory.class.php
@@ -412,7 +412,19 @@ class PluginFusioninventoryInventoryComputerInventory {
          }
          $inputdb['rules_id'] = $data['_ruleid'];
          $inputdb['method'] = 'inventory';
-         $pfIgnoredimportdevice->add($inputdb);
+         $inputdb['plugin_fusioninventory_agents_id'] = $_SESSION['plugin_fusioninventory_agents_id'];
+
+         // if existing ignored device, update it
+         if ($found = $pfIgnoredimportdevice->find("`plugin_fusioninventory_agents_id` =
+                                                    '".$inputdb['plugin_fusioninventory_agents_id']."'",
+                                                   "`date` DESC",
+                                                   1)) {
+            $agent         = array_pop($found);
+            $inputdb['id'] = $agent['id'];
+            $pfIgnoredimportdevice->update($inputdb);
+         } else {
+            $pfIgnoredimportdevice->add($inputdb);
+         }
       }
    }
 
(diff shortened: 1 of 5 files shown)


inc/inventorycomputerinventory.class.php
```diff
@@ -412,7 +412,18 @@ function sendCriteria($p_DEVICEID, $arrayinventory) {
          }
          $inputdb['rules_id'] = $data['_ruleid'];
          $inputdb['method'] = 'inventory';
-         $pfIgnoredimportdevice->add($inputdb);
+         $inputdb['plugin_fusioninventory_agents_id'] = $_SESSION['plugin_fusioninventory_agents_id'];
+
+         // if existing ignored device, update it
+         if ($found = $pfIgnoredimportdevice->find("`plugin_fusioninventory_agents_id` =
```
ddurieux: problem because `$inputdb['serial']` not exist in some cases (case the device not have serial number or have serial number blacklisted)

inc/inventorycomputerinventory.class.php
```diff
@@ -412,7 +412,18 @@ function sendCriteria($p_DEVICEID, $arrayinventory) {
...
+                                                    '".$inputdb['plugin_fusioninventory_agents_id']."'
+                                                    AND `serial`='".$inputdb['serial']."'")) {
```
ddurieux: use find to get only 1 item, not need get all, so the query will be faster with a `limit 1`

install/update.php
```diff
@@ -2128,12 +2128,18 @@ function do_ignoredimport_migration($migration) {
...
+   $a_table['fields']['plugin_fusioninventory_agents_id']
+                                    = array('type'    => 'integer',
+                                            'value'   => NULL);
 
    $a_table['oldfields']  = array();
 
    $a_table['renamefields'] = array();
 
    $a_table['keys']   = array();
+   $a_table['keys'][] = array('field' => 'plugin_fusioninventory_agents_id',
```
ddurieux: you search with plugin_fusioninventory_agents_id and serial but add only a simple index instead double index 
(plugin_fusioninventory_agents_id + serial)

orthagh:
I removed serial search from original commit (no need for changing index so).
I don't see the point to search on it (i'll ask to @wawax when he will be back).

I also added order and limit to the find.

PR #2048, opened 2017-02-02
add link to computer in agent log

orthagh:
Ex:

image


orthagh:
The modification is done into taskjobs.js not taskjobs.min.js
I could push also in this file, but i think we need to check this behavior (maybe doing the minified file only in release archive ?)

ddurieux:
you're right, not think about that :(

PR #2051, opened 2017-02-02
closes #2023 Add new checks : skip, info, warning

### LINKED ISSUE #2023: Improve audit for packages
As described on the agent side here : https://github.com/fusioninventory/fusioninventory-agent/issues/205 

We need to improve the audit actions.
Add new actions : 

- Skip job
- Warning
- Info

The server should check the deploy task version. If the task doesn't handle the new parameters, it should return "ignore" instead.
A new checkbox must be displayed to send the check output or not back to the server.

A a new audit called "sha512 file is not" to the audit list.

wawax:
Work in progress


ddurieux:
if it's in wip, can you add label wip or add `[wip]` in title of the PR?

PR #2057, opened 2017-02-07
retrieve force start button in task (only for debug mode)

orthagh:
Also cleaned PluginFusioninventoryTaskjob::forceRunningTask who seems to be a dead code.

### DIFF
--- a/inc/task.view.class.php
+++ b/inc/task.view.class.php
@@ -522,6 +522,44 @@ class PluginFusioninventoryTaskView extends PluginFusioninventoryCommonView {
       return true;
    }
 
+   function showFormButtons($options = array()) {
+      if (isset($this->fields['id'])) {
+         $ID = $this->fields['id'];
+      } else {
+         $ID = 1;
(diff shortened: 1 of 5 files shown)


inc/task.view.class.php
```diff
@@ -522,6 +522,44 @@ function showForm($id, $options=array()) {
       return true;
    }
 
+   function showFormButtons($options = array()) {
+      if (isset($this->fields['id'])) {
+         $ID = $this->fields['id'];
+      } else {
+         $ID = 1;
```
trasher: I do not understand why "1" here.

orthagh: hum, no idea for the moment.
I need to test the suppression of this part.

orthagh: Hum, it's a copy paste from https://github.com/glpi-project/glpi/blob/master/inc/commondbtm.class.php#L2092
I think for the case of fi tasks, we can clean this else

PR #2058, opened 2017-02-07
use cache on dynamic group

orthagh:
When demanding a b/collect/index.php?action=getJobs , 
All agent are computed again and so a dynamic group listing launch.
Depending on the criteria saved in this group, query could be very long and executed again for **each** agent of this group (imagine a deployment for 1000 computers querying the dynamic group each). 

If i understand well, the code check if an agent always belongs to a prepared job and cancel if not.
For mass deployment, it's pretty aggressive on the server.

I suggest, in this pr, to cache computers_id of dynamic groups at the preparation of the task (use_cache = false, the default), and at the moment of execution, use this cache (use_cache = true).

Maybe, we can do better for this problem, but i don't see atm.

Also cleaned some incoherence in update.php

### DIFF
--- a/inc/deploygroup_dynamicdata.class.php
+++ b/inc/deploygroup_dynamicdata.class.php
@@ -190,27 +190,66 @@ class PluginFusioninventoryDeployGroup_Dynamicdata extends CommonDBChild {
     *
     * @since 0.85+1.0
     *
-    * @param object $group PluginFusioninventoryDeployGroup instance
-    * @return array
+    * @param group the group object
+    * @param use_cache retrieve computers_id from cache (computers_id_cache field)
+    * @return an array of computer ids
     */
-   static function getTargetsByGroup(PluginFusioninventoryDeployGroup $group) {
-      $search_params = PluginFusioninventoryDeployGroup::getSearchParamsAsAnArray($group, false,true);
-      if (isset($search_params['metacriteria']) && empty($search_params['metacriteria'])) {
-         unset($search_params['metacriteria']);
-      }
-      $search_params['sort'] = '';
-
-      //Only retrieve computers IDs
-      $results = Search::prepareDatasForSearch('PluginFusioninventoryComputer', $search_params, array('2'));
-      Search::constructSQL($results);
-      $results['sql']['search'] = str_replace("`mainitemtype` = 'PluginFusioninventoryComputer'",
-              "`mainitemtype` = 'Computer'", $results['sql']['search']);
-      Search::constructDatas($results);
-
+   static function getTargetsByGroup(PluginFusioninventoryDeployGroup $group, $use_cache = false) {
       $ids = array();
-      foreach ($results['data']['rows'] as $row) {
-         $ids[$row['id']] = $row['id'];
+
+      if (!$use_cache || !$ids = self::retrieveCache($group)) {
+         $search_params = PluginFusioninventoryDeployGroup::getSearchParamsAsAnArray($group, false,true);
+         if (isset($search_params['metacriteria']) && empty($search_params['metacriteria'])) {
+            unset($search_params['metacriteria']);
+         }
+
+         //force no sort (Search engine will sort by id) for better performance
+         $search_params['sort'] = '';
+
+         //Only retrieve computers IDs
+         $results = self::getDatas(
+            'PluginFusioninventoryComputer',
+            $search_params,
+            array('2')
+         );
+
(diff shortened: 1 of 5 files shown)


inc/deploygroup_dynamicdata.class.php
```diff
@@ -190,27 +190,66 @@ static function getDatas($itemtype, $params, array $forcedisplay=array()) {
...
-   static function getTargetsByGroup(PluginFusioninventoryDeployGroup $group) {
-      $search_params = PluginFusioninventoryDeployGroup::getSearchParamsAsAnArray($group, false,true);
-      if (isset($search_params['metacriteria']) && empty($search_params['metacriteria'])) {
-         unset($search_params['metacriteria']);
-      }
-      $search_params['sort'] = '';
+   static function getTargetsByGroup(PluginFusioninventoryDeployGroup $group, $use_cache = false) {
+      $ids = array();
+
+      if (!$use_cache || !$ids = self::retrieveCache($group)) {
```
ddurieux: Not really agree with that, if you call the function with `$use_cache` = 'false' and you have ids in cache, it use cache.
So 2 options:
* if we not use cache, really not use it
* keep like it but remove the param `$use_cache` in the function

install/mysql/plugin_fusioninventory-empty.sql
```diff
@@ -915,6 +915,7 @@ CREATE TABLE `glpi_plugin_fusioninventory_deploygroups_dynamicdatas` (
   `plugin_fusioninventory_deploygroups_id` int(11) NOT NULL DEFAULT '0',
   `fields_array` text DEFAULT NULL,
   `can_update_group` tinyint(1) NOT NULL DEFAULT '0',
+  `computers_id_cache` text DEFAULT NULL,
```
ddurieux: On big computers database, perhaps use `longtext` or `blob`

trasher: OK, but, should not `blob` cause issues (I'm thinking about future multi-database support mainly)?

ddurieux: I think about that too, but postgresql use `bytea` instead `blob` and the `text` is unlimited in size instead MySQL, so in all cases, will need modify it :p 
I preferer `longtext` now

trasher: 👍 for longtext

inc/deploygroup_dynamicdata.class.php
orthagh: maybe the notation is a little too compact but here is the different cases:

- $use_cache = false, the first part of the 'if' will trigger (!$use_cache) and the second part will not (due to the '|' usage)
- $use_cache = true, the first part return false, so we try with the second part to load the cache:
  * it succeeds: we skip the block
  * it fails: we execute the block and load data from Search engine.

For resume, as the mentionned 'if' have an OR condition, if the first part is evaluated as true, the second part will not be evaluated.

trasher: I agree, the second part will not be called if `$use_cache` is set to false.

inc/deploygroup_dynamicdata.class.php
```diff
@@ -190,27 +190,66 @@ static function getDatas($itemtype, $params, array $forcedisplay=array()) {
...
+         }
+
+         //store results in cache (for reusing on agent communication)
+         self::storeCache($group, $ids);
+      }
+
+      return $ids;
+   }
+
+   static function storeCache(PluginFusioninventoryDeployGroup $group, $ids) {
```
trasher: Add phpdoc

inc/deploygroup_dynamicdata.class.php
```diff
@@ -190,27 +190,66 @@ static function getDatas($itemtype, $params, array $forcedisplay=array()) {
...
+      global $DB;
+
+      $query = "UPDATE ".self::getTable()."
+                SET `computers_id_cache` = '".$DB->escape(json_encode($ids))."'
+                WHERE `plugin_fusioninventory_deploygroups_id` = '".$group->getID()."'";
+      return $DB->query($query);
+   }
+
+   static function retrieveCache(PluginFusioninventoryDeployGroup $group) {
```
trasher: Add phpdoc

orthagh:
Any news on this PR.
Did you see my [answer](https://github.com/fusioninventory/fusioninventory-for-glpi/pull/2058#discussion_r111165378) above ?

ddurieux:
Ok, I'm ok with that, you can merge ;)

orthagh:
Thank you !

PR #2065, opened 2017-02-13
Fix delete of audits and files for a job

wawax:
This PR fixes audits and files deletion for a package.


wawax:
Not relevant anymore, since these bugs have been fixed in the audits PR that has been approved by @trasher today.

Issue #230, opened 2015-09-29
Many bugs on inventory in task


ddurieux:
---

Author Name: **David Durieux** (@ddurieux)
Original Redmine Issue: 631, http://forge.fusioninventory.org/issues/631
Original Date: 2011-02-18
Original Assignee: David Durieux

---

See : http://forum.fusioninventory.org/viewtopic.php?id=220

PR #2066, opened 2017-02-14
closes #2023 Add new return values for audit checks: skip, info, warn…

### LINKED ISSUE #2023: Improve audit for packages

wawax:
Rework audit display to be more comprehensible : describe, in a human readable form the purpose of the audit
Add new audit returns : skip job, report info, report warning


g-bougard:
fusioninventory-agent needs Deploy task v2.2 provided with [deploy audit enhancement PR #230](https://github.com/fusioninventory/fusioninventory-agent/pull/230)

PR #2082, opened 2017-02-27
Add code to refresh the plugin's homepage

wawax:
Refresh plugin's homepage.
This is controlled by the user's preferences to refresh tickets list.


front/menu.php
```diff
@@ -50,6 +50,15 @@
    Html::header(__('FusionInventory', 'fusioninventory'), $_SERVER["PHP_SELF"], "plugins",
                 "pluginfusioninventorymenu", "menu");
 
+   //TODO : in GLPI 9.2, to replace by Html::manageRefreshPage()
```
trasher: Maybe just test the presence of the new method? Well, not very important ;)

PR #2083, opened 2017-02-27
Add computer's hostid as the agent now retrieves it

wawax:
Retrieve and display computers hostid.
This information may be used by software vendors for licensing purpose.


hook.php
```diff
@@ -2433,8 +2438,20 @@ function postShowtab($params) {
                    );
 
                    echo "</td>";
+
```
trasher: I do not understand why teh /tr? Was it missing?

PR #2097, opened 2017-03-09
Add rule import/export header

wawax:
Make it possible to import/export rules for FusionInventory rules engine.

Related to PR https://github.com/glpi-project/glpi/pull/1931 for GLPI


trasher:
If you rely on a recent commit on GLPI side; we'll have to change minimal required version to 9.1.3; and the next FI release cannot be done until GLPI 9.1.3has been released as well.

Note that currently, GLPI version in the bugfixes branch is still 9.1.2.

wawax:
Hi @trasher ,

yes you're right, I'll rebase on the master.

trasher:
Tests are failing because FI use removed methods from the core :/

PR #2113, opened 2017-03-30
GLPI default branch has changed

trasher:


trasher:
Finally, only one branch is used on each FI branches... Is it useful to keep that `if`? We could set only one `$GLPI` variable and use it directly in the checkout (using `git clone -b $GLPI [...]`).

PR #2118, opened 2017-04-04
Continue to improve software deployment audits

wawax:
Review audits for Windows registry : 
- change labels
- add new audits from checking registry value existance & type

Add a "debug" tab for packages when user is in debug mode.

Add unit tests for deployment audits


wawax:
rebased and replace by https://github.com/fusioninventory/fusioninventory-for-glpi/pull/2160

PR #2131, opened 2017-04-12
Clean repository on package purge

orthagh:


orthagh:
will do a test for this

orthagh:
Hum, in fact, i don't see any way to achieve a fileupload with phpunit.
So if no suggestions, i think we can merge as if

trasher:
As far as I can see in existing code, nothing would prevent to simulate an upload.

The only thing that should prevent fake file upload would be use of `is_uploaded_file` or 'move_uploaded_file` methods. If those functions are not used, you just have to cal the upload method with required informations you can set from unit test.

inc/deployfile.class.php
```diff
@@ -667,29 +667,38 @@ static function remove_item($params) {
...
          if ($checked >= "1" || $checked == "on") {
             //get sha512
-            $sha512 = $datas['jobs']['associatedFiles'][$index];
+            $sha512 = $data['jobs']['associatedFiles'][$index];
 
             //remove file
             unset($files[$index]);
-            //array_splice($datas['jobs']['associatedFiles'], $index, 1);
-            unset($datas['associatedFiles'][$sha512]);
+            //array_splice($data['jobs']['associatedFiles'], $index, 1);
```
trasher: Maybe this line could be dropped

inc/deployfile.class.php
```diff
@@ -1056,36 +1066,53 @@ static function addFileInRepo($params) {
     * Remove file from the repository
     *
     * @param string $sha512 sha512 of the file
-    * @param integer $packages_id id of the package
     * @return boolean
     */
-   static function removeFileInRepo($sha512, $packages_id) {
+   static function removeFileInRepo($sha512) {
+      $repoPath        = GLPI_PLUGIN_DOC_DIR."/fusioninventory/files/repository/";
```
trasher: Should not paths be declared elsewhere? I guess same ones are used on the upload part; it would be better to have only one var to change/adapt for both of them.

phpunit/2_Integration/Deploy/RepositoryTest.php
```diff
@@ -0,0 +1,194 @@
...
+
+   @package   FusionInventory
+   @author    Alexandre Delaunay
+   @co-author
+   @copyright Copyright (c) 2010-2017 FusionInventory team
+   @license   AGPL License 3.0 or (at your option) any later version
+              http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   @link      http://www.fusioninventory.org/
+   @link      http://forge.fusioninventory.org/projects/fusioninventory-for-glpi/
+   @since     2013
```
trasher: Bad year :p

phpunit/2_Integration/Deploy/RepositoryTest.php
```diff
@@ -0,0 +1,194 @@
...
+
+   ------------------------------------------------------------------------
+ */
+
+class RepositoryTest extends RestoreDatabase_TestCase {
```
trasher: If I've understand well the above code, it is possible to have directories that will be dropped if there is no longer any file into. Maybe add a tests case for that?

PR #2137, opened 2017-04-26
add includeoldjobs params, and fix net task display

orthagh:


inc/task.class.php
```diff
@@ -1006,21 +1006,24 @@ function getJoblogs($task_ids = array()) {
 
          foreach ($targets as $target) {
             $item_type = key($target);
-            $item_id = current($target);
+            $item_id   = current($target);
+            $item_name = "";
+            if (strpos($item_id, '$#$')) {
```
trasher: What if strpos() === 0?

orthagh: good point

ajax/taskjob_logs.php
```diff
@@ -50,13 +50,8 @@
    Session::checkCentralAccess();
 }
 
-//unlock session since access checks have been done
-session_write_close();
```
ddurieux: You remove that, so if have many logs, it will locks all php pages loaded

orthagh: ok for that, i'll move to later (to permits to write my vars)

orthagh: done

ddurieux: Ok, parfait, un petit squash et on merge ;)

PR #2140, opened 2017-04-27
export jobs

orthagh:


inc/taskview.class.php
```diff
@@ -478,6 +516,8 @@ public function submitForm($postvars) {
          }
          $this->update($postvars);
          Html::back();
+      } else if (isset($postvars['export_jobs'])) {
+         $this->csvExport($postvars);
```
ddurieux: In task.form.php, you call this function `submitForm` before the checkright, so you need add a checkright here in the case of export

inc/task.class.php
```diff
@@ -1625,6 +1631,152 @@ function post_updateItem($history=1) {
...
+      $agent_state_types = ['prepared', 'cancelled', 'running', 'success', 'error' ];
+      if (isset($params['agent_state_types'])) {
+         $agent_state_types = $params['agent_state_types'];
+      }
+
+      // 0 : no debug (really export to csv,
+      // 1 : display final table,
+      // 2 : also display json
+      define('DEBUG_CSV', 0);
+      if (!DEBUG_CSV) {
```
ddurieux: You define `DEBUG_CSV` 2 lines before so I think it's for debug and need to be remove, but see that, I'm not sure it right defined in another part of code

inc/task.class.php
```diff
@@ -1625,6 +1631,152 @@ function post_updateItem($history=1) {
...
+            }
+         }
+      }
+
+      // clean old temporary variables
+      unset($task, $job, $target, $agent);
+
+      if (!DEBUG_CSV) {
+         define('SEP', ';');
+         define('NL', "\r\n");
```
ddurieux: Why not get the global config `csv separator` of GLPI ?

PR #2142, opened 2017-05-02
fix fatal error 'Call to undefined method PluginFusioninventoryConfig…

orthagh:
showTabs function is removed from glpi master


orthagh:
It seems we have some additional occurence of this showTabs function in code.
I will complete the PR soon

ddurieux:
Yes, I think it require more modifications ;)

orthagh:
closed in favor of #2143

PR #2145, opened 2017-05-03
review dashboard (no more table positioning)

orthagh:


trasher:
With Firefox:
- before
board_orig-fullpage
- after
board_rewrite-fullpage

orthagh:
Fixed

trasher:
I still have the issue with your latest commits and browser cache cleaned :/

css/views.css
```diff
@@ -838,11 +838,51 @@ table.runs td.comment {
...
 
 .copyleft {
   display:inline-block;
   transform: rotate(180deg);
-}
\ No newline at end of file
+}
+
+.fi_board {
+   width: 1300px;
```
trasher: Why this value? It's not compliant with "small" screens and its breaks responsiveness, does'nt it?

PR #2149, opened 2017-05-10
Previous OS stuff are still used; fixes #2102

### LINKED ISSUE #2102: [FI 9.1+1.1] Operating system information display N/A in Locks (fields) tab
Hi,
In the "**Locks (fields)**" tab for a computer, values of last inventory for _Operating system, OS version and Service pack_ display **N/A**.
Feature "**Manage operating system name**" is enable in the configuration of the plugin.
fi_os-bug1

It seems these values use the old fields _operatingsystemversions_id, operatingsystemservicepacks_id and operatingsystems_id_ in "**Import information**" tab who are at **-1** instead of the new values in _plugin_fusioninventory_computeroperatingsystems_id_.
fi_os-bug2

So when you update one or more lock fields, "**Operating System**" tab loose these 3 values.
We must force an inventory on the computer or wait for an automatic to retrieve correct information.
Thanks.

trasher:
This is to fix issue spotted in #2102; but not only.

Setting `['Computer']['operatingsystem{XYZ}']` to 0 would cause issues elsewhere, since this reference is still used (see for example ESX detection a few lines below OS stuff in Formatconvert class).

### DIFF
--- a/inc/formatconvert.class.php
+++ b/inc/formatconvert.class.php
@@ -568,9 +568,24 @@ class PluginFusioninventoryFormatconvert {
             $array_tmp['operatingsystemservicepacks_id'] = '';
          }
          $a_inventory['fusioninventorycomputer']['plugin_fusioninventory_computeroperatingsystems_id'] = $array_tmp;
-         $a_inventory['Computer']['operatingsystemversions_id'] = 0;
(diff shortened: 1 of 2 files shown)


trasher:
Thanks :)

PR #2151, opened 2017-05-10
filter actor dropdown in tasks

orthagh:
The dropdown will show only agents or computers which have the module enabled (or in global conf).
I tested on a db with > 10k agents, i had a tiny lag to display the second dropdown (<500ms).

Before:
image

After:
image

### DIFF
--- a/inc/taskjobview.class.php
+++ b/inc/taskjobview.class.php
@@ -395,12 +395,15 @@ class PluginFusioninventoryTaskjobView extends PluginFusioninventoryCommonView {
     * @param array $options
     */
    public function ajaxModuleItemsDropdown($options) {
+      global $DB;
+
       $moduletype = $options['moduletype'];
-      $itemtype = $options['itemtype'];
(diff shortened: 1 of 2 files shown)


inc/taskjobview.class.php
```diff
@@ -411,13 +414,56 @@ public function ajaxModuleItemsDropdown($options) {
             break;
 
       }
+
+      // filter actor list with active agent and with current module active
+      $condition = "";
+      if ($moduletype == "actors"
+          && in_array($itemtype, ["Computer", "PluginFusioninventoryAgent"])) {
+         // remove install suffix from deploy
+         $modulename = str_replace('DEPLOYINSTALL', 'DEPLOY', strtoupper($method));
```
ddurieux: DEPLOYINSTALL no longer exist because it has been updated with migration script in 9.1+1.0 version

ddurieux: It's ok, I have not read right :( sorry...

PR #2156, opened 2017-05-11
Deploy debug tab

wawax:
Do not display debug informations below the package definition, but in a specific tab instead.
this tab only appears when user is in debug mode.


wawax:
Well I'd remove the save button to prevent changing the package json, and then break the internal structure.
Have you already used this feature ?

ddurieux:
Seems you're the only one yes :D

PR #2160, opened 2017-05-22
Improve audits & checks

wawax:
Checks : 
- add a warning message to recommand or require agent version 2.3.20 or higher
- add registry key/value syntax examples
- improve unit tests to reflect these new changes

Actions : 
- add a label for an action
- add a new parameter (already describe in the specs) to indicate the number of line of the command output to retrieve
- unit tests


inc/deploycheck.class.php
```diff
@@ -226,14 +263,17 @@ static function displayList(PluginFusioninventoryDeployPackage $package, $datas,
...
-         if (is_numeric($check['value'])
-            && in_array($check['type'], ['freespaceGreater', 'fileSizeLower',
-                                         'fileSizeEquals', 'fileSizeGreater'])) {
-            if ($check['type'] == "freespaceGreater") {
+         switch ($check['type']) {
+            case 'freespaceGreater':
                $check['value'] = $check['value'] * 1024 * 1024;
-            }
-            $check['value'] = PluginFusioninventoryDeployFile::processFilesize($check['value']);
+               break;
```
orthagh: i'm not sure about this ```break;```.

Before, the ```1024 * 1024``` part was executed also with ``processFilesize`` part in case of freespaceGreater

inc/deploycheck.class.php
```diff
@@ -226,14 +263,17 @@ static function displayList(PluginFusioninventoryDeployPackage $package, $datas,
...
+            case 'fileSizeLower':
+            case 'fileSizeGreater':
+            case 'fileSizeEquals':
+               $check['value'] = PluginFusioninventoryDeployFile::processFilesize($check['value']);
+               break;
+            default :
```
orthagh: not usefull, remove the default part

inc/deploycheck.class.php
```diff
@@ -403,24 +462,39 @@ static function getValues($type, $data, $mode) {
...
-         $mandatory_mark = '';
-      }
+      $mandatory_mark = ($mandatory?self::getMandatoryMark():'');
 
       switch ($check_type) {
          case "winkeyExists":
          case "winkeyMissing":
-            $values['path_label']  = __("Key", 'fusioninventory').$mandatory_mark;
-            $values['value_label'] = FALSE;
+         $values['path_label']         = __("Path to the key", 'fusioninventory').$mandatory_mark;
```
orthagh: indentation seems ko here

inc/deploycheck.class.php
```diff
@@ -513,6 +595,14 @@ static function displayAjaxValues($config, $request_data, $rand, $mode) {
                   $values['value']."' /></td>";
                break;
 
+            case 'registry_type':
+               echo "<td>";
+               self::dropdownRegistryTypes($values['value']);
+               echo "</td>";
+               break;
+
+            case "input+unit":
```
orthagh: duplicate line

orthagh:
LGTM, maybe a second eye, @trasher   ?

inc/deploycheck.class.php
```diff
@@ -96,12 +108,12 @@ static function getLabelForAType($type) {
...
     */
    static function getUnitLabel() {
-      return array(
-         "B"  => __("B", 'fusioninventory'),
-         "KB" => __("KiB", 'fusioninventory'),
-         "MB" => __("MiB", 'fusioninventory'),
-         "GB" => __("GiB", 'fusioninventory')
-      );
+      return [
+               "B"  => __("B", 'fusioninventory'),
```
trasher: Would be better to use same units as in the core (Kio, Mio and so on). Maybe use directly Toolbox::getSize() instead of a local method?

phpunit/custom-build.xml
```diff
@@ -9,7 +9,7 @@
   <target name="phpunit.unit.custom" depends='plugin-fusioninventory.prepare' description="Run unit tests with PHPUnit">
    <phpunit.run type="custom" configuration='${fusioninventory.directory}/phpunit/testsuites_integration.xml'>
      <testfile>
-       <arg path='${tests.directory}/2_Integration/Deploy/RepositoryTest.php' />
```
trasher: This should probably not be commited ;)

trasher:
OK for me. Please rebase and squash before we can merge ;)

PR #2164, opened 2017-06-01
Improve drive devices detection

wawax:
Rework drive device detection method.
Also add unit tests for it.

### DIFF
--- a/inc/formatconvert.class.php
+++ b/inc/formatconvert.class.php
@@ -2143,8 +2143,8 @@ class PluginFusioninventoryFormatconvert {
     * @return string "Drive" or "HardDrive"
     */
    static function getTypeDrive($data) {
-      $to_match_regex = ['rom', 'dvd', 'blue.{0,1}ray', 'reader',
-                         'sd.{0,1}card', 'micro{0,1}sd', 'mmc'];
+      $to_match_regex = ['rom', 'dvd', 'blu[\s-]*ray', 'reader',
+                         'sd[\s-]*card', 'micro[\s-]*sd', 'mmc'];
       $found_drive    = false;
       foreach ($to_match_regex as $regex) {
          foreach (['TYPE', 'MODEL', 'NAME'] as $field) {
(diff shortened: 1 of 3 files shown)


inc/formatconvert.class.php
```diff
@@ -2310,21 +2143,22 @@ static function printerInventoryTransformation($array) {
...
-               OR (preg_match("/blue.{0, 1}ray/i", $data["MODEL"]))))
-            OR
-         ((isset($data['NAME'])) AND
-              ((preg_match("/rom/i", $data["NAME"])) OR (preg_match("/dvd/i", $data["NAME"]))
-               OR (preg_match("/blue.{0, 1}ray/i", $data["NAME"]))))) {
-
-         return "Drive";
-      } else {
-         return "HardDrive";
+      $to_match_regex = ['rom', 'dvd', 'blue.{0,1}ray', 'reader',
```
trasher: I do not know what the agent will give as data; but the name is blu-ray, not blue-ray.

I guess you should use something like `(\s-)?` (not 100% sure, would need to be tested, but you got the idea) rather than `{0, 1}` to handle multiple chars cases.

phpunit/1_Unit/FormatConvertDataTest.php
```diff
@@ -359,4 +359,48 @@ function testComputerConvert() {
...
+                             ['MODEL' => '', 'NAME' => 'ST3250824AS Q'],
+                             ['TYPE'  => 'sd  card'],
+                             ['TYPE'  => 'blue  ray'],
+                             ['TYPE' => 'DISK',
+                              'MODEL' => 'PM951NVMe SAMSUNG 256GB',
+                              'NAME' => 'nvme0n1']
+                           ];
+
+
+      foreach ($tests as $return_expected => $values) {
```
trasher: Use a dataProvider maybe?

phpunit/custom-build.xml
```diff
@@ -9,7 +9,7 @@
   <target name="phpunit.unit.custom" depends='plugin-fusioninventory.prepare' description="Run unit tests with PHPUnit">
    <phpunit.run type="custom" configuration='${fusioninventory.directory}/phpunit/testsuites_integration.xml'>
      <testfile>
-       <arg path='${tests.directory}/1_Unit/Deploy/DeployactionTest.php' />
```
trasher: Useless ;)

PR #2169, opened 2017-06-05
Import IP ranges and their related SNMP authentication from a CSV file

mohierf:


mohierf:
@ddurieux : it looks okay 😉

PR #2172, opened 2017-06-08
 Store global counter of printer in GLPI core closes #678

### LINKED ISSUE #678: Store global counter of printer in GLPI core
---

Author Name: **David Durieux** (@ddurieux)
Original Redmine Issue: 1305, http://forge.fusioninventory.org/issues/1305
Original Date: 2011-10-27

---

See glpi ticket : https://forge.indepnet.net/issues/3205

wawax:
Update GLPI last page counter when fusioninventory performs a remote SNMP inventory

### DIFF
--- a/phpunit/1_Unit/PrinterUpdateTest.php
+++ b/phpunit/1_Unit/PrinterUpdateTest.php
@@ -155,7 +155,7 @@ class PrinterUpdate extends RestoreDatabase_TestCase {
          'is_template'          => '0',
          'template_name'        => NULL,
          'init_pages_counter'   => '0',
-         'last_pages_counter'   => '0',
+         'last_pages_counter'   => 15134,
          'users_id'             => '0',
          'groups_id'            => '0',
          'states_id'            => '0',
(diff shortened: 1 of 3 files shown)


phpunit/1_Unit/PrinterUpdateTest.php
```diff
@@ -166,7 +166,28 @@ public function PrinterGeneral() {
...
 
+      $DB->connect();
+
+      $printer = new Printer();
+      $printer->getFromDB(1);
+      //Counter has been updated
+      $this->assertEquals($printer->fields['last_pages_counter'], 15134);
+
+      //No log has been added in the printer's historical
+      $nb = countElementsInTable('glpi_logs',
```
trasher: I do not understand why you check logs here. You just issue a getFromDB; logs will never be updated.
I guess you should look at that, but where the printer is really added in the database.

phpunit/1_Unit/PrinterUpdateTest.php
```diff
@@ -166,7 +166,28 @@ public function PrinterGeneral() {
       $this->assertEquals($a_reference, $printer->fields);
    }
 
+   /**
+    * @test
+    */
+   public function printerGlpiCounter() {
```
trasher: This test seems to depend on another one; it should be specified

PR #2175, opened 2017-06-09
Software system category

wawax:
Import software category from the agent into glpi


trasher:
Need https://github.com/glpi-project/glpi/pull/2003 to be merged first

PR #2189, opened 2017-06-20
Fix tasks_id when preparing a on demand deploy job

wawax:


inc/deploypackage.class.php
```diff
@@ -1660,7 +1660,8 @@ function deployToComputer($computers_id, $packages_id, $users_id) {
             $pfTaskJob->update($input);
 
             //Store the task's ID
-            $tasks_id = $data['id'];
```
orthagh: I agree the previous line was wrong, but i think you don't need to fetch more the db to get the task id.
As i see below in the code, we fetch all columns of job table, so in ``$data`` we should have ``plugin_fusioninventory_tasks_id`` field.

orthagh:
On your last commit, you did some changes in unit tests, i think there are unrelated

PR #2198, opened 2017-07-03
User interactions

wawax:
Factorize audit, file and action code : a new PluginFusioninventoryPackageItem abstract class has been introduced.
Modify some static methods to non static in order to do some factorization and overriding.

Add code to support userinteractions (agent version 2.3.21 mandatory for this new feature).
Added unit tests for user interactions.


orthagh:
closed in favor of #2366

PR #2205, opened 2017-07-05
Only restrict by entity if by_entites boolean is set to true

wawax:


inc/task.class.php
```diff
@@ -1519,7 +1519,8 @@ static function getItemsFromDB($filter) {
 
       // Filter by entity
       if (isset($filter['by_entities'])
-              && is_bool($filter['by_entities'])) {
+	      && is_bool($filter['by_entities'])
```
trasher: What about just `(bool)$filter['by_entities']`?

PR #2206, opened 2017-07-05
Deploymirror migration and searchoption

wawax:
Enable existing mirror by default during migration.
Add search options for deploymirror.is_active


ddurieux:
I could be nice to check if the field is right added in DB in tests (in upgrade tests)

PR #2227, opened 2017-07-29
[WIP] feature inventory import from JSON format

btry:
JSON is converted into XML then processed


btry:
Hi

The purpose of this PR was to permit an agent to send an inventory using JSON format. 

I need to ckech with @g-bougard and the Flyve MDM team if this is still relevant.

PR #2233, opened 2017-08-02
unclean_cross_side_scripting_deep of commande

lermit4:


wawax:
Hi @lermit4 
Thanks for the PR. Could you provide a use case where the action is not correclty managed, so I can reproduce the issue ?

lermit4:
echo "write in file" >> c:\TEMP\HELLO.txt
the ">" is replaced by '&lt;', '&gt;' in the command line and windows or linux doesn't like this

wawax:
thank you. I reproduce your issue !

wawax:
thank you. I reproduce your issue !

ddurieux:
I have moved your code in another file + add a test.
Thanks for your contribution and sorry for the too long delay (https://github.com/fusioninventory/fusioninventory-for-glpi/pull/2532)

PR #2235, opened 2017-08-03
Import printer port during snmp inventory if ifnumber is defined

wawax:
Import a printer port if ifnumber is defined (and not the opposite)Import a printer port if ifnumber is defined (and not the opposite)


inc/communicationnetworkdiscovery.class.php
```diff
@@ -257,7 +257,7 @@ function sendCriteria($arrayinventory) {
          $_SESSION['plugin_fusinvsnmp_taskjoblog']['comment'] = '==importdenied== '.
                                                                   implode(", ", $a_text);
          $this->addtaskjoblog();
-
+
```
ddurieux: Remove white spaces please ;)

PR #2242, opened 2017-08-16
Add default status for snmp devices

wawax:


inc/communicationnetworkdiscovery.class.php
```diff
@@ -333,6 +333,10 @@ function rulepassed($items_id, $itemtype, $entities_id=0) {
          $input = array();
          $input['date_mod'] = date("Y-m-d H:i:s");
          $input['entities_id'] = $entities_id;
+
+         //Add defaut status if there's one defined in the configuration
+         $input = PluginFusioninventoryToolbox::addDefaultStateIfNeeded('snmp', $input);
```
ddurieux: Not sure it's a good idea to manage the status in the discovery, for the Inventory yes, but not really agree with Discovery because it's not a very `secure` information

PR #2247, opened 2017-08-26
Clean import infos tab : drop inventory array for debug (useless), pu…

wawax:
Clean the import information tab : 

- remove useless inventory PHP array
- add a clean form to download the XML inventory file
- add a pager to browse rules


wawax:
@ddurieux what do you think about @trasher 's proposal to drop serialized_inventory ? We both agree that the display was useless, and it seems that the field is not used except for display purpose.
Could you confirm that ?

Thanks.

ddurieux:
We use the serialized for the lock form

PR #2251, opened 2017-08-29
Remove dead code

wawax:
Remove dead code

- automatic selection of agent has been remove from task management
- there's no snmp dictionnary mechanism in the agent : drop the code on the server side
- clean unsued methods in task management


wawax:
replaced by https://github.com/fusioninventory/fusioninventory-for-glpi/pull/2394

PR #2274, opened 2017-09-20
Networkequipment firmwares

trasher:


inc/formatconvert.class.php
```diff
@@ -1935,6 +1927,42 @@ static function networkequipmentInventoryTransformation($array) {
...
+      if (isset($array['INFO']['FIRMWARE']) || isset($array['FIRMWARES'])) {
+         $a_firmwares = [];
+         $mapping = [
+            'DESCRIPTION'  => 'description',
+            'MANUFACTURER' => 'manufacturers_id',
+            'NAME'         => 'name',
+            'TYPE'         => 'devicefirmwaretypes_id',
+            'VERSION'      => 'version'
+         ];
+         if (isset($array['FIRMWARES'])) {
```
wawax: I would suggest, if FIRMWARES is not an array, to create and array and manage only one case

PR #2314, opened 2017-10-18
Update behavior to permit display task log when the task is not active

ddurieux:
When a task is disabled, there no way to see executions reports.


mohierf:
Please do not merge this. I will include this modificaton in my next PR and it will obviously conflict :(

PR #2316, opened 2017-10-19
Uniform fusioninventory group label in task job

ddurieux:


inc/staticmisc.class.php
```diff
@@ -157,7 +157,7 @@ static function getModulesMethods() {
    static function task_definitiontype_wakeonlan($a_itemtype) {
 
       $a_itemtype['Computer'] = Computer::getTypeName();
-      $a_itemtype['PluginFusioninventoryDeployGroup'] = __('Dynamic Group');
+      $a_itemtype['PluginFusioninventoryDeployGroup'] = __('Fusioninventory group');
```
orthagh: 'Fusioninventory group' is not in GLPI locales context, so you should add ``, 'fusioninventory'``

wawax: hi @ddurieux ! can you perform the requested change ? thanks

wawax: I suggest to modify the locale in DeployGroup::getTypeName and use this method instead of calling directly gettext

PR #2323, opened 2017-10-23
purge agent on computer purge when no task associated

orthagh:
Following the irc discussion

### DIFF
--- a/hook.php
+++ b/hook.php
@@ -1920,30 +1920,41 @@ function plugin_pre_item_update_fusioninventory($parm) {
 /**
  * Manage pre-item purge an item
  *
- * @global object $DB
  * @param object $parm
  * @return object
  */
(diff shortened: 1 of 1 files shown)


trasher:
Tests are failing ;)

ddurieux:
I'm not agree, delete the taskjobstates based on agents_id can be very long and it's not a good idea to make it here. Perhaps add a cron task to clean this the night

orthagh:
we don't delete taskjobstates, we delete agent if no taskjobstates exist for it

ddurieux:
Ok, you're right, but use limit 1 in `$states = $pfTaskjobstate->find(...` it will prevent get all lines when have very too many ;)

orthagh:
ok, good point, i push a pr this morning for that

PR #2328, opened 2017-10-26
Fix delete all collect information of all computer when send data to …

ddurieux:
…an agent


orthagh:
closed in favor of #2354

PR #2346, opened 2017-11-15
rework taskjobs counters

orthagh:
Ok, THESE jobs counters bother me from long time and today i went with shovel to remove all complexity.

I think, let me know if i'm wrong, the counters should "resume" the last state of an agent.
If, in the log, we have for the same agent:
- job 2 -> prepared
- job 1 -> error 
The counter `In error` should not count this agent, only `Prepared` (and `Not done yet` in cons)

So now, only the last state influence the counters, we remove **older** states before adding a new one

Nota: this PR is very readable with side by side diff, not at all in unified


orthagh:
Close this one, will do a better fix

PR #2352, opened 2017-11-17
Location rule engine for snmp asset

stonebuzz:
add capability to location rule engine to work with snmp asset


Phill93:
How do i use this feature?

ddurieux:
What information you don't find it?

Phill93:
The option to use this rule. The assignment of location based on snmp devices

PR #2354, opened 2017-11-20
run collect shouldnt delete old content for other agents

orthagh:
I dropped some parts of collect which delete old content because:
- it deletes all data (also for others computers)
- it's redundant with another parts of code (ex for registry: https://github.com/fusioninventory/fusioninventory-for-glpi/blob/master/inc/collect_registry_content.class.php#L177)

For files collect, the [updateComputer](https://github.com/fusioninventory/fusioninventory-for-glpi/blob/master/inc/collect_file_content.class.php#L149-L159) part doesn't have the deletion process, so i kept the query, but i added the filter on computers_id to avoid data deletion for others computers

### DIFF
--- a/inc/collect.class.php
+++ b/inc/collect.class.php
@@ -581,10 +581,6 @@ class PluginFusioninventoryCollect extends CommonDBTM {
                    'uuid'     => $taskjobstate->fields['uniqid'],
                    '_sid'     => $reg['id']
                );
-               //clean old registries results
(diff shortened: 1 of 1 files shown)


orthagh:
lol, i just found that i fixed the same thing as #2328 .
I think the part of file is better here

trasher:
Unit tests are failing

orthagh:
fixed

PR #2356, opened 2017-11-22
Drop os id from comparison to prevent migration related issues

trasher:
The issue: OS id was not correctly set previously. On the first inventory on a 9.2, this is fixed.... But all software are removed and reinstalled.


trasher:
@ddurieux I'd like your opinion on this issue... If that seems OK to you, I'll check why tests are failing; I do not want to spend time on it if this is not ok ;)

wawax:
I close this PR as it has been replaced by https://github.com/fusioninventory/fusioninventory-for-glpi/pull/2437

Issue #2368, opened 2017-12-02
Transifex access


AnaelMobilia:
Hello,

I wanted to update some typos on the frenc language file, but i cant' have access to https://www.transifex.com/ddurieux/teams/797

Is it a way to have access granted ?

Regards,
Anael

ddurieux:
Hello, you can request here: https://www.transifex.com/ddurieux/FusionInventory/content/

ddurieux:
Oh sorry, yes it's the right url you give

AnaelMobilia:
Thanks ! :)

PR #2370, opened 2017-12-03
Markdown + update additionnal ressources

AnaelMobilia:
asciidoc -> markdown
Update old URL (forge) + add transifex link


AnaelMobilia:
Add transifex link for #2368

PR #2372, opened 2017-12-04
remove agent tab in computer, and get agent status in auto_inventory block

orthagh:
Status is called by ajax, so the computer page can't stuck.
I change showInfoForComputer prototype to pass Computer object needed by ``showRemoteStatus`` method

screenshot: 
image

### DIFF
--- a/ajax/remote_status.php
+++ b/ajax/remote_status.php
@@ -1,0 +1,112 @@
+<?php
+
+/*
+   ------------------------------------------------------------------------
+   FusionInventory
+   Copyright (C) 2010-2011 by the FusionInventory Development Team.
+
+   http://www.fusioninventory.org/   http://forge.fusioninventory.org/
+   ------------------------------------------------------------------------
+
+   LICENSE
+
+   This file is part of FusionInventory project.
+
+   FusionInventory is free software: you can redistribute it and/or modify
+   it under the terms of the GNU Affero General Public License as published by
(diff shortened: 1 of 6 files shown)


trasher:
The URL call that previously blocks the page is now called from ajax, this should no longer blocks page load.

orthagh:
Hello @ddurieux, like @trasher said, we manage the blocking request by calling it in **ajax** (after dom ready)
So on page display (or reload), we don't stuck waiting for answer

ajax/remote_status.php
```diff
@@ -0,0 +1,90 @@
...
+      switch($agentStatus['message']) {
+
+         case 'executing scheduled tasks':
+         case 'running':
+            $agentStatus['message'] = __('Running');
+            break;
+
+         case 'noanswer':
+            $agentStatus['message'] = "<i class='fa fa-exclamation'></i>".
+                                      __('Impossible to communicate with agent!', 'fusioninventory');
```
wawax: I would suggest to change the locale to "Impossible to reach the agent" or "Cannot contact the agent"

orthagh:
@ddurieux, i did better.

The status is still in main form, but not auto triggered, it now needs a click to send the first request, see:

- status ko:
peek 05-12-2017 11-21

- status ok:
peek 05-12-2017 11-23

ddurieux:
Ok for me ;)

PR #2373, opened 2017-12-04
deploy module should be enabled by default

orthagh:
As collect (on by default like inventory) and to easing deploy usage, module for deploy should be enabled by default on plugin installation.

### DIFF
--- a/install/install.php
+++ b/install/install.php
@@ -294,7 +294,7 @@ function pluginFusioninventoryInstall($version, $migrationname='Migration') {
 
       $input = array();
       $input['modulename'] = "DEPLOY";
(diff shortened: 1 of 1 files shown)


ddurieux:
Not really agree, collect and inventory only read, with deploy you write, so I prefer keep it disabled by default

ddurieux:
@wawax and @trasher, what's your opinion about that ?

trasher:
Sounds like many users do use deploy feature (or maybe have issues with it :D)... So it should make sense to enable it by default.
What could be the implications? Does it cause any possible issue even when not used? Or performance issues maybe?

PR #2394, opened 2017-12-27
Remove dead code in tasks. Translation strings that are not yet trans…

wawax:
- Clean unused methods
- Use constants for tasks related states
- Translate hardcoded strings in tasks logs

### DIFF
--- a/inc/communicationnetworkdiscovery.class.php
+++ b/inc/communicationnetworkdiscovery.class.php
@@ -76,7 +76,7 @@ class PluginFusioninventoryCommunicationNetworkDiscovery {
       if (isset($a_CONTENT['PROCESSNUMBER'])) {
          $_SESSION['glpi_plugin_fusioninventory_processnumber'] = $a_CONTENT['PROCESSNUMBER'];
          if ($pfTaskjobstate->getFromDB($a_CONTENT['PROCESSNUMBER'])) {
-            if ($pfTaskjobstate->fields['state'] != "3") {
+            if ($pfTaskjobstate->fields['state'] != PluginFusioninventoryTaskjobstate::FINISHED) {
                $pfTaskjobstate->changeStatus($a_CONTENT['PROCESSNUMBER'], 2);
                if ((!isset($a_CONTENT['AGENT']['START']))
                        AND (!isset($a_CONTENT['AGENT']['END']))) {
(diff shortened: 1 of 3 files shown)


inc/communicationnetworkdiscovery.class.php
```diff
@@ -103,45 +103,32 @@ function import($p_DEVICEID, $a_CONTENT, $arrayinventory) {
       }
 
       if ($pfTaskjobstate->getFromDB($a_CONTENT['PROCESSNUMBER'])) {
-         if ($pfTaskjobstate->fields['state'] != "3") {
+         if ($pfTaskjobstate->fields['state'] != PluginFusioninventoryTaskjobstate::FINISHED) {
             $pfImportExport = new PluginFusioninventorySnmpmodelImportExport();
             $errors .= $pfImportExport->import_netdiscovery($a_CONTENT, $p_DEVICEID);
             if (isset($a_CONTENT['AGENT']['END'])) {
-               if ((isset($a_CONTENT['DICO'])) AND ($a_CONTENT['DICO'] == "REQUEST")) {
```
trasher: I really do not know how to be sure this is still used

inc/taskjob.class.php
```diff
@@ -530,15 +447,9 @@ function dropdownAction($myname, $actiontype, $method, $actiontypeid, $value=0,
...
    function getAgents($module) {
-
-      $array = [];
-      $array[".1"] = " ".__('Auto managenement dynamic of agents', 'fusioninventory');
-
-      $array[".2"] = " ".__('Auto managenement dynamic of agents (same subnet)', 'fusioninventory');
-
+      $array         = [];
       $pfAgentmodule = new PluginFusioninventoryAgentmodule();
-      $array1 = $pfAgentmodule->getAgentsCanDo(strtoupper($module));
```
trasher: Hum... Maybe the original way was better for perfs

inc/communicationnetworkdiscovery.class.php
wawax: it was an old behavior of the agent:it had an auto update mechanism for network discovery mapping file. If the agent had a outdated version, the server would send a message to stop the process. Next time, the server would send the new file to the agent and the process could continue.
If was complicated and frustrating for the user, it has been removed a long time ago.

trasher: OK, thank you for the explanations :)

inc/taskjob.class.php
wawax: you're right I've reverted this modification

inc/taskjoblog.class.php
```diff
@@ -899,40 +686,33 @@ function getDivState($state, $type='div') {
...
 
-         case 2:
+         case self::TASK_OK:
             return "<".$type." style='background-color: rgb(0, 255, 0);-moz-border-radius: 4px;".
                      "-webkit-border-radius: 4px;-o-border-radius: 4px;padding: 2px;' ".
                      "align='center' width='".$width."'>".
-                     "<strong>".__('Ok', 'fusioninventory')."</strong></".$type.">";
+                     "<strong>".__('OK')."</strong></".$type.">";
 
-         case 3:
```
trasher: I guess this is also something that is no longer used now?

inc/taskjob.class.php
```diff
@@ -530,19 +447,18 @@ function dropdownAction($myname, $actiontype, $method, $actiontypeid, $value=0,
     * @return array [id integed agent id] => $name value agent name
     */
    function getAgents($module) {
-
-      $array = [];
-      $array[".1"] = " ".__('Auto managenement dynamic of agents', 'fusioninventory');
```
trasher: Another old thing not longer used?

wawax: yes, it was an old task feature that has been deprecated during the task rework

inc/taskjoblog.class.php
wawax: yes, this taskjoblog state has never been used...

PR #2395, opened 2017-12-29
Coding standards using PHPCS

wawax:
The code has been rewritten by phpcs using GLPI's coding standard rules
There's no code change instead of coding standard


trasher:
You should add a CS check in the travis file ;)

PR #2397, opened 2018-01-01
Fix #2398 - Dynamic computers groups search result

### LINKED ISSUE #2398: Dynamic computer groups - search result do not include all the fields
In the Associated items tab, only the computers name is displayed.

When searching computers on several fields (eg. name, IP, location, ...) the search fields are not displayed in the result list

mohierf:
Fix #2398 and add the matching items count on the Associated items tab.

Include a search preview under the search form for a preview of the search result.

This makes it much more comfortable to view the results of a search 😉

### DIFF
--- a/inc/deploygroup_dynamicdata.class.php
+++ b/inc/deploygroup_dynamicdata.class.php
@@ -126,7 +126,7 @@ class PluginFusioninventoryDeployGroup_Dynamicdata extends CommonDBChild {
       if (isset($params['metacriteria']) && !is_array($params['metacriteria'])) {
          $params['metacriteria'] = [];
       }
-      $params['target'] = PluginFusioninventoryDeployGroup::getFormURLWithID($item->getID(), true);
+      $params['target'] = PluginFusioninventoryDeployGroup::getSearchEngineTargetURL($_GET['id'], true);
 
       $data = Search::prepareDatasForSearch('PluginFusioninventoryComputer', $params);
       Search::constructSQL($data);
@@ -178,7 +178,7 @@ class PluginFusioninventoryDeployGroup_Dynamicdata extends CommonDBChild {
                $params['metacriteria'] = [];
             }
 
-            $params['target'] = PluginFusioninventoryDeployGroup::getFormURLWithID($item->getID(), true);
+            $params['target'] = PluginFusioninventoryDeployGroup::getSearchEngineTargetURL($_GET['id'], true);
             self::showList('PluginFusioninventoryComputer', $params, array('1', '2'));
             return TRUE;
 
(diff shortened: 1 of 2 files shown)


inc/deploygroup_dynamicdata.class.php
```diff
@@ -88,13 +88,60 @@ function getTabNameForItem(CommonGLPI $item, $withtemplate=0) {
...
+   /**
+    * Get the count of items matching the dynamic search criteria
+    *
+    * @param object $item the item object
+    * @param integer $withtemplate 1 if is a template form
+    * @return string name of the tab
+    */
+   function getMatchingItemsCount(CommonGLPI $item) {
+
+      $params_dyn = array();
```
wawax: please use [] for array definition instead of array()

inc/deploygroup_dynamicdata.class.php
```diff
@@ -88,13 +88,60 @@ function getTabNameForItem(CommonGLPI $item, $withtemplate=0) {
...
+      }
+      $params = PluginFusioninventoryDeployGroup::getSearchParamsAsAnArray($item, false);
+      $params['massiveactionparams']['extraparams']['id'] = $_GET['id'];
+      foreach ($params_dyn as $key => $value) {
+         $params[$key] = $value;
+      }
+      if (isset($params['metacriteria']) && !is_array($params['metacriteria'])) {
+         $params['metacriteria'] = array();
+      }
+      $params['target'] = Toolbox::getItemTypeFormURL("PluginFusioninventoryDeployGroup" , true).
```
wawax: Please use PluginFusioninventoryDeployGroup::getFormURLWithID instead of Toolbox::getItemTypeFormURL

inc/deploygroup_dynamicdata.class.php
mohierf: Ok

inc/deploygroup_dynamicdata.class.php
mohierf: Ok

mohierf: Note that this is not the only piece of code where this is used ...

wawax: in this case I suggest to factorize this code if possible  :)

mohierf:
@wawax : I fixed according to your request. I may squash the commits and rebase before you merge this?

trasher:
I agree; searches may be quite long and take many resources.

mohierf:
@wawax 
My tests conditions:
 - database containing **8500** computers.
 - requests such as: 'name' contains and 'IP address' contains and 'user' is ...

And I got very good response time (< 1 second)

ddurieux:
Search with operatingsystem and software and software version (and the best, test with computer AND NOT software name contain xxx), I think it will be more slowly.
With fusion group, in general cases, you have big queries

mohierf:
Some more tests:
1/ name contains, AND IP contains, AND Software/Name contains: 0.10 seconds
2/ name contains, AND IP contains, AND Software/Name contains AND Software/Version name contains: 0.10 seconds
3/ name contains, AND IP contains, AND OS/comments contains, AND Software/Name contains AND Software/Version name contains: 0.10 seconds


All the queries I just tested have a response time near 0.1 second !

mohierf:
I just changed the title because I intend to make one more commit on this PR today...

PR #2405, opened 2018-01-03
Tasks jobs management

### LINKED ISSUE #2403: Jobs management in the tasks - only one job is enough
As discussed with @ddurieux only one job should be managed in the tasks. The main reason is that several jobs are not correctly managed when the task is running ... I do not really know why 😉

### LINKED ISSUE #2404: Computers massive actions - Target a task
This feature is not correctly implemented currently. Several jobs are created in the tasks and created jobs are not updated when needed ...

mohierf:
Fix #2403: only one job per task
Fix #2404: correctly target a task with computers massive actions

### DIFF
--- a/inc/task.class.php
+++ b/inc/task.class.php
@@ -2007,15 +2007,31 @@ class PluginFusioninventoryTask extends PluginFusioninventoryTaskView {
             break;
 
          case 'target_task':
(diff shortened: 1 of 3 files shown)


mohierf:
@wawax : the last build failed ... and I cannot restart the build on Travis for this project ... any way to restart ?

PR #2408, opened 2018-01-04
Closes #2407 - add a configuration parameter to configure default job…

### LINKED ISSUE #2407: Allow to configure a default value for re-preparing a job
Add a configuration parameter to define if successful executed jobs must be re-prepared once finished.

This parameter will be used for a new task creation to set if jobs must be re-prepared when they are executed.

mohierf:
…s re-preparation behavior

### DIFF
--- a/inc/config.class.php
+++ b/inc/config.class.php
@@ -153,6 +153,8 @@ class PluginFusioninventoryConfig extends CommonDBTM {
       $input['manage_osname']         = 1;
       $input['clean_on_demand_tasks'] = -1;
 
+      $input['reprepare_job']         = 0;
+
       if (!$getOnly) {
          $this->addValues($input);
       }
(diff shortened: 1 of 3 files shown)


inc/taskview.class.php
```diff
@@ -484,6 +484,7 @@ public function submitForm($postvars) {
     */
    function getEmpty() {
       parent::getEmpty();
-      $this->fields['reprepare_if_successful'] = 1;
+      $pfConfig = new PluginFusioninventoryConfig();
+      $this->fields['reprepare_if_successful'] = $pfConfig->getValue('reprepare_job');;
```
wawax: there's 2 ;

mohierf: Oops! 😉

inc/config.class.php
```diff
@@ -375,7 +377,11 @@ function showForm($options=[]) {
       echo "<td>";
       Dropdown::showYesNo("extradebug", $this->isFieldActive('extradebug'));
       echo "</td>";
-      echo "<td colspan=2></td>";
+
+      echo "<td>".__('Re-prepare jobs once finished', 'fusioninventory')."</td>";
```
wawax: I would say 'By default, re-prepare successfull jobs'. What do you think about that ?

mohierf: I would say: 'Re-prepare successfull jobs'. The 'By default' amkes no sense because the check box nera the message indicates whether it is set or not

wawax: ok let's go for Re-prepare successfull jobs !

mohierf: Updated and squashed

PR #2410, opened 2018-01-04
[WiP] Includes fields plugin in the search criteria

mohierf:
This intends to allow using the Fields plugin defined fields in the search criteria for the computers groups

**Note** there is still a huge problem to cope with because of the search itemtype that is using 'PluginFusioninventoryComputer' which is not known by the Fields plugin when the SQL request is built :/


wawax:
Hi,
how does the search behave if the fields plugin gets disabled ? Are they any risk that the search returns much more entries than needed ?

inc/search.class.php
```diff
@@ -88,15 +88,21 @@ class PluginFusioninventorySearch extends CommonDBTM {
    /**
     * Cloned Core function to display with our require.
     *
+    * !!! This function is never used anywhere in the plugin !!!
```
trasher: It should be removed if not used (I do not see any usage); but it may be in plugins that depends on FI.

mohierf: I confirmed with @ddurieux that this function has been used but is now deprecated. I simple commented to raise an alert before removing 😉 but I will remove this code.

mohierf:
@wawax : the Fields plugin search fields are only included if the Fields plugin is installed and enabled. As such, no impact at all if the plugin is not present...

inc/search.class.php
trasher: You can also mark it as deprecated in the phpodc (using '@deprecated`) so this can be removed later. As you prefer ;)

wawax:
an example: your dynamic group targets all computers having the value X in the Y field (brought by the Fields plugin). For one reason the plugin is disabled, I suppose that the condition is ignored, so all computers will be targeted by the group.
This is, I suppose, not what the administrator wanted ^ ^

mohierf:
@trasher : sorry for the .idea files ... I will remove

mohierf:
Sure @wawax ... this case may be surprising but, if the group includes some Fields plugin information, it should be ignored rather than used to run a task. 

Note that this PR is not currently fully functional because I have a problem to manage correctly the request mixing some FusionInventoryComputer objects and the Fields pluginthat is expecting some Computer objects :/

mohierf:
I am closing this pull request because it will be quite complex to merge. I will open a new cleaner one ...

PR #2418, opened 2018-01-05
Use fields plugin in the search engine of computers groups

mohierf:
This allow to use the Fields plugin defined fields in the search criteria of the static or dynamic computers groups. Same for the computer information collected in registry or WMI ...

**Note** that this PR is based on the `add-itesm-count` branch (the 4 first commits) because it will be easier to merge once https://github.com/fusioninventory/fusioninventory-for-glpi/pull/2397 is merged 😉  You may even merge this PR and forget the other one.


mohierf:
@wawax : would you mind I squash the commits? My idea is to have only 2/3 commits in this PR...

inc/deploygroup_dynamicdata.class.php
```diff
@@ -154,13 +241,18 @@ static function showList($itemtype, $params, $forcedisplay) {
       $_GET['_in_modal'] = true;
```
wawax: it doesn't seems to be a good idea to set the variable to true because the GLPI header is lost when browsing search results

mohierf: Ok. You are right ! It is obviously a bad idea! I do not even understand how this should run correctly on my test instance 😞

PR #2419, opened 2018-01-05
Update local and manual unit test procedure

mohierf:


phpunit/README.asc
```diff
@@ -68,24 +68,80 @@ mysql> FLUSH PRIVILEGES;
...
+    Load Done...
+
 
 == Usage
 
 This set of testsuites are meant to be executed with +ant+.
 
+To install Apache ant :
+
+1. sudo apt-get install ant
```
trasher: This will not work on many systems... Better to add  link to the ant website and say something more generic

mohierf:
Once again sorry for the .idea :)
I will update the .gitignore ...

phpunit/README.asc
mohierf: Ok

PR #2425, opened 2018-01-10
Update ru_RU.po

GubanovAnton:


trasher:
Locales are handled on Transifex. Thank you.

Issue #2412, opened 2018-01-04
Plugin main page is very long to display


mohierf:
Currently, on my test server the plugin main page need **almost 30 seconds** to get displayed.

I found out that this is because of a very huge and not optimized SQL query built by the `Task::getJobLogs` function.

The SQL request:
```
SELECT task.`id` AS 'task.id', task.`name` AS 'task.name', job.`id` AS 'job.id', job.`name` AS 'job.name', job.`method` AS 'job.method', agent.`id` AS 'agent.id', agent.`name` AS 'agent.name', agent.`computers_id` AS 'agent.computers_id', run.`id` AS 'run.id', run.`itemtype` AS 'run.itemtype', run.`items_id` AS 'run.items_id', run.`state` AS 'run.state', log.`date` AS 'log.last_date', UNIX_TIMESTAMP(log.`date`) AS 'log.last_timestamp', log.`id` AS 'log.last_id', log.`comment` AS 'log.last_comment'
FROM `glpi_plugin_fusioninventory_agents` AS agent
INNER JOIN ( SELECT MAX(run.`id`) AS max_id, run.`plugin_fusioninventory_agents_id`, run.`plugin_fusioninventory_taskjobs_id`, run.`items_id`, run.`itemtype`, MAX(log.`id`) AS max_log_id
FROM ( SELECT *
FROM ( SELECT *, @num := IF(@agent_id = plugin_fusioninventory_agents_id && @taskjob_id = plugin_fusioninventory_taskjobs_id && @items_id = items_id && @itemtype = itemtype, @num:= @num + 1, 1) AS row_num, @agent_id:=plugin_fusioninventory_agents_id AS tmp_var1, @taskjob_id:=plugin_fusioninventory_taskjobs_id AS tmp_var2, @items_id:=items_id AS tmp_var3, @itemtype:=itemtype AS tmp_var4
FROM glpi_plugin_fusioninventory_taskjobstates
ORDER BY plugin_fusioninventory_taskjobs_id, plugin_fusioninventory_agents_id, itemtype, items_id, id DESC ) AS limited_states
WHERE row_num <= 2 ) AS run
[...]

PR #2427, opened 2018-01-10
For #2412 - temporarily remove the deployment chart on plugin home page

mohierf:


mohierf:
@wawax : I found a solution to improve the huge request that breaks the plugin home page. I will submit a PR on the same branch as the one you just merged...

I informed @ddurieux and @trasher this morning on the IRC 😉

PR #2455, opened 2018-01-26
Check if port exists using logical number first and then mac address

wawax:


inc/inventorycommon.class.php
```diff
@@ -139,8 +154,8 @@ function importPorts($itemtype, $a_inventory, $items_id) {
             $pfNetworkPort->update($a_port);
          } else {
             // Update port
-            $networkports_id = $a_ports_DB['id'];
-            $a_port['id'] = $a_ports_DB['id'];
+            $networkports_id = $networkPort->getID();
```
trasher: Maybe it should be better not to use `getID()`; we've faced some strange issues using it.

wawax: done

phpunit/custom-build.xml
```diff
@@ -9,7 +9,7 @@
   <target name="phpunit.unit.custom" depends='plugin-fusioninventory.prepare' description="Run unit tests with PHPUnit">
    <phpunit.run type="custom" configuration='${fusioninventory.directory}/phpunit/testsuites_integration.xml'>
      <testfile>
-       <arg path='${tests.directory}/2_Integration/SoftwareVersionUpdateTest.php' />
+       <arg path='${tests.directory}/1_Unit/PrinterUpdateTest.php' />
```
trasher: You should avoid changes on this file... :/

Maybe add it to your local excludes (`echo "phpunit/custom-build.xml" >> .git/info/exclude`)

PR #2458, opened 2018-01-26
Feature/release script

trasher:


trasher:
One thing I've missed is to replace the development flag. Need to fix directory and archives names; and this should be ok :)

trasher:
TODOs has been done :)

The release script is adapted from GLPI one; since what is provided in glpi-project/tools only works with semver.

wawax:
LGTM.
Just a simple question at the beginning of the script to ask the admin if the transifex sync has been recently made.

tools/make_release.sh
```diff
@@ -0,0 +1,132 @@
...
+fi
+
+echo "Check XML WF"
+if ! xmllint --noout fusioninventory.xml; then
+   echo "XML is *NOT* well formed. Exiting."
+   exit 1;
+fi
+
+echo "Retrieve PHP vendor"
+composer install --no-dev --optimize-autoloader --prefer-dist --quiet
```
ddurieux: perhaps delete forlder `vendor` if exist before do this?

tools/make_release.sh
```diff
@@ -0,0 +1,132 @@
...
+
+echo "Set version and official release"
+sed \
+   -e 's/"PLUGIN_FUSIONINVENTORY_OFFICIAL_RELEASE", "0"/"PLUGIN_FUSIONINVENTORY_OFFICIAL_RELEASE", "0"/' \
```
ddurieux: You need set to 1 :p

tools/make_release.sh
```diff
@@ -0,0 +1,132 @@
...
+   -i setup.php
+
+echo "Minify stylesheets and javascripts"
+$INIT_PWD/vendor/bin/robo minify
+
+echo "Compile locale files"
+./tools/update_mo.pl
+
+echo "Delete various scripts and directories"
+\rm -rf vendor;
```
ddurieux: We need keep `vendor` to deliver with the release

tools/make_release.sh
trasher: It does not exists, since the directory in /tmp is created from `git checkout-index` ;)

tools/make_release.sh
trasher: Well catched :D

tools/make_release.sh
trasher: Currently, the vendor directory only contains development/release related libs

PR #2462, opened 2018-01-31
Do not add entries in the log at asset creation

wawax:
Harmonize log behavior at asset creation based on what's done for computers : 
- do not add a log entry during OS creation for a computer
- do not add log entries when manually inserting a XML of SNMP inventory for printers & network devices

### DIFF
--- a/inc/inventorynetworkequipmentlib.class.php
+++ b/inc/inventorynetworkequipmentlib.class.php
@@ -123,7 +123,8 @@ class PluginFusioninventoryInventoryNetworkEquipmentLib extends PluginFusioninve
          'FROM'  => getTableForItemType("PluginFusioninventoryNetworkEquipment"),
          'WHERE' => ['networkequipments_id' => $items_id]
       ];
-      foreach ($DB->request($params) as $data) {
+      $iterator = $DB->request($params);
+      while ($data = $iterator->next()) {
          foreach ($data as $key=>$value) {
             $db_networkequipment[$key] = Toolbox::addslashes_deep($value);
          }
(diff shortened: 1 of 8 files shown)


inc/inventorynetworkequipmentlib.class.php
```diff
@@ -117,11 +118,12 @@ function updateNetworkEquipment($a_inventory, $items_id) {
...
-         FROM `".  getTableForItemType("PluginFusioninventoryNetworkEquipment")."`
-         WHERE `networkequipments_id` = '$items_id'";
-      $result = $DB->query($query);
-      while ($data = $DB->fetch_assoc($result)) {
+
+      $params = [
+         'FROM'  => getTableForItemType("PluginFusioninventoryNetworkEquipment"),
+         'WHERE' => ['networkequipments_id' => $items_id]
+      ];
+      foreach ($DB->request($params) as $data) {
```
trasher: Maybe avoid calling a method in the loop. I'm not sure a `foreach` is OK here; you should have used somtehing like `while ($data = $iterator->next())`

inc/inventoryprinterlib.class.php
```diff
@@ -102,16 +102,16 @@ function updatePrinter($a_inventory, $printers_id) {
...
-      $query = "SELECT *
-            FROM `".  getTableForItemType("PluginFusioninventoryPrinter")."`
-            WHERE `printers_id` = '$printers_id'";
-      $result = $DB->query($query);
-      while ($data = $DB->fetch_assoc($result)) {
+      $params = [
+         'FROM'  => getTableForItemType("PluginFusioninventoryPrinter"),
+         'WHERE' => ['printers_id' => $printers_id]
+      ];
+      foreach ($DB->request($params) as $data) {
```
trasher: Same here

PR #2469, opened 2018-02-03
Download xml file & change name of the file

wawax:
Factorize code to display FI infos on assets
Rename XML files stored on the filesystem : add a .xml suffix
Add a download XML function for network equipments and printers

### DIFF
--- a/inc/item.class.php
+++ b/inc/item.class.php
@@ -1,0 +1,302 @@
+<?php
+
+/**
+ * FusionInventory
+ *
+ * Copyright (C) 2010-2016 by the FusionInventory Development Team.
+ *
+ * http://www.fusioninventory.org/
+ * https://github.com/fusioninventory/fusioninventory-for-glpi
+ * http://forge.fusioninventory.org/
+ *
+ * ------------------------------------------------------------------------
+ *
+ * LICENSE
+ *
+ * This file is part of FusionInventory project.
+ *
+ * FusionInventory is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * FusionInventory is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with FusionInventory. If not, see <http://www.gnu.org/licenses/>.
+ *
+ * ------------------------------------------------------------------------
+ *
+ * This file is used to manage the extended information of a computer.
+ *
+ * ------------------------------------------------------------------------
+ *
+ * @package   FusionInventory
+ * @author    David Durieux
+ * @copyright Copyright (c) 2010-2016 FusionInventory team
+ * @license   AGPL License 3.0 or (at your option) any later version
+ *            http://www.gnu.org/licenses/agpl-3.0-standalone.html
+ * @link      http://www.fusioninventory.org/
+ * @link      https://github.com/fusioninventory/fusioninventory-for-glpi
+ *
+ */
+
+if (!defined('GLPI_ROOT')) {
+   die("Sorry. You can't access directly to this file");
+}
+
+/**
+ * Common class to manage informations FI display on an asset
+ */
+class PluginFusioninventoryItem extends CommonDBTM {
+
+   /**
+    * The right name for this class
+    *
+    * @var string
+    */
+   static $rightname = '';
+
+   /**
+   * The itemtype we are processing
+   */
+   public $itemtype  = '';
+
+   /**
+   * Get automatic inventory info for a computer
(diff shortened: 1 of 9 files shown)


inc/item.class.php
```diff
@@ -0,0 +1,239 @@
...
+      return true;
+   }
+
+   /**
+   * Add more info to the tab if needed
+   * @since 9.2+2.0
+   *
+   * @param array $options display options
+   */
+   function addMoreInfos($options = []) {
```
trasher: Why is this method empty?

inc/item.class.php
```diff
@@ -0,0 +1,239 @@
...
+
+   }
+
+   /**
+   * Get a FI class instance representing an itemtype
+   * @since 9.2+2.0
+   *
+   * @param array $options display options
+   */
+   static function getFIItemClassInstance($itemtype) {
```
trasher: Maybe should we add a message in logs when it is called with an unknown item type?

inc/networkequipment.class.php
```diff
@@ -90,24 +92,26 @@ static function displayTabContentForItem(CommonGLPI $item, $tabnum=1, $withtempl
...
-                 array('target'=>$CFG_GLPI['root_doc'].
-                                    '/plugins/fusioninventory/front/switch_info.form.php'));
+            case 'NetworkPort$1':
+               $pfNetworkEquipment->showExtendedInfos($item, $options);
+               break;
          }
-         return TRUE;
       }
-      return FALSE;
+      return false;
```
trasher: Method will always return `false`; is that intended?

inc/networkequipment.class.php
```diff
@@ -153,11 +156,11 @@ function showForm(CommonDBTM $item, $options=array()) {
          }
       }
 
-      if (isset($_POST['displaysnmpinfo'])) {
-         $this->showNetworkEquipmentInformation($id, $options);
-         return;
-      }
-//$_SESSION['plugin_fusioninventory_networkportview'] = 'glpi';
+      //if (isset($_POST['displaysnmpinfo'])) {
```
trasher: Please drop commented code

inc/networkequipment.class.php
```diff
@@ -431,8 +432,8 @@ class='tab_bg_1_2'>".$icon.$item->getLink(1);
                   $link1 = $item->getLink(1);
                   $link = str_replace($item->getName(0), $NetworkPort->fields["mac"],
                                       $item->getLink());
-//                  $link2 = str_replace($item->getName(0), $NetworkPort->fields["ip"],
-//                                       $item->getLink());
+                  //                  $link2 = str_replace($item->getName(0), $NetworkPort->fields["ip"],
```
trasher: Please drop commented code

inc/networkequipment.class.php
```diff
@@ -489,14 +489,14 @@ function updateNetworkInfo($id, $configsecurities_id, $sysdescr) {
...
 
-
    /**
     * Display network equipment extended information
     *
     * @global object $DB
     * @param object $item
     * @param array $options
     */
+    /*
```
trasher: Should be dropped

inc/item.class.php
```diff
@@ -0,0 +1,239 @@
...
+
+   /**
+   * The itemtype we are processing
+   */
+   public $itemtype  = '';
+
+   /**
+   * Get automatic inventory info for a computer
+   * @since 9.1+1.2
+   * @param $items_id the item ID to look for
```
trasher: Param type is missing

inc/item.class.php
```diff
@@ -0,0 +1,239 @@
...
+         return $this->fields;
+      } else {
+         return [];
+      }
+   }
+
+   /**
+   * Form to download the XML inventory file
+   *
+   * @param $items_id the item ID to look for
```
trasher: Param type is missing
Return is missing as well

inc/item.class.php
```diff
@@ -0,0 +1,239 @@
...
+         echo "&nbsp;$message</a>";
+         echo "</th></tr></table>";
+      }
+   }
+
+   /**
+    * Display form
+    *
+    * @param object $item CommonDBTM instance
+    * @param array $options
```
trasher: Param description is missing. Params/types/description should be indented the same

inc/networkequipment.class.php
wawax: done

inc/networkequipment.class.php
wawax: done

inc/networkequipment.class.php
wawax: done

inc/networkequipment.class.php
wawax: no it should return true, as for other objects

inc/item.class.php
wawax: this method enables a child class to add more data to display. By default it's empty, PluginFusioninventoryNetworkEquipment overrides the method

inc/item.class.php
wawax: done

inc/item.class.php
wawax: done

inc/item.class.php
wawax: done

PR #2477, opened 2018-02-07
Extra debug log for collected information

mohierf:
Clean and add some extra debug log for rules engine - this will make it easier to understand some problems...


inc/inventoryrulelocationcollection.class.php
```diff
@@ -95,10 +95,6 @@ function getTitle() {
     * @return array
     */
    function prepareInputDataForProcess($input, $params) {
```
wawax: if there's nothing in this method now, why not to remove it ?

mohierf: Ok

ddurieux:
Squashed and merged manually, thanks

PR #2533, opened 2018-05-02
Naturally sort network ports

po1vo:
Couldn't find a way to do the job with just SQL, so i brought PHP to help.


ddurieux:
Thanks for the contribution ;)

Issue #2553, opened 2018-05-28
Error on upgrade 


camlafit:
Hello

In dev instance (glpi 9.3 RC1 and fusioninventory glpi took from master : 1d3ea42530f5d0253b24a71a0775551d61233471)

``` 
PHP Warning: rename(/var/www/glpi/files/_plugins/fusinvdeploy/files,/var/www/glpi/files/_plugins/fusioninventory/files): Directory not empty in /var/www/glpi/plugins/fusioninventory/install/update.php at line 296

Fatal error: Uncaught RuntimeException: Cannot run an UPDATE query without WHERE clause! in /var/www/glpi/inc/dbmysql.class.php:973 Stack trace: #0 /var/www/glpi/inc/dbmysql.class.php(1001): DBmysql->buildUpdate('glpi_plugin_fus...', Array, Array) #1 /var/www/glpi/plugins/fusioninventory/install/update.php(1613): DBmysql->update('glpi_plugin_fus...', Array, Array) #2 /var/www/glpi/plugins/fusioninventory/install/update.php(364): do_entities_migration(Object(Migration)) #3 /var/www/glpi/plugins/fusioninventory/hook.php(825): pluginFusioninventoryUpdate('9.2+1.0', 'Migration') #4 /var/www/glpi/inc/plugin.class.php(701): plugin_fusioninventory_install() #5 /var/www/glpi/front/plugin.form.php(47): Plugin->install('3') #6 {main} thrown in /var/www/glpi/inc/dbmysql.class.php on line 97
```

camlafit:
Ok  understood 

glpi project reject any update without any where condition : 
https://github.com/glpi-project/glpi/blob/9.3-RC1/inc/dbmysql.class.php#L972

Wolvverine:
now from master branch:

2018-06-05T11:15:49.044204566Z 10.0.0.67 - - [05/Jun/2018:11:15:49 +0000] "POST /front/plugin.form.php HTTP/1.1" 200 25841 "http://docker:80/front/plugin.php" "Mozilla/5.0 (Windows NT 6.3; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/66.0.3359.181 Safari/537.36"
[...]

Issue #2544, opened 2018-05-08
After Upgrade plugi to version 9.2 +2.0 version not update


ermannog:
Hello
I have upgrade the version of plugin from 9.2 +1.0 to 9.2 +2.0 but in the list of plugins in GLPI aways show 9.1 +1.0 as version...

Is possible correct this manually?

ddurieux:
not normal, perhaps cache of the broswer and perhaps restart php-fpm

ermannog:
I open the table glpi_plugins in the DB and in the column version there is 9.2 +1.0 and not 9.2 +2.0...

ddurieux:
So the plugin is not upgraded :p

ermannog:
Yes but I retry to:
disable
delete the folder of the plugin
copy again the folder of the verion 9.2 +2.0
but the option for update the plugin not appear..only Activate and Unistall..

ddurieux:
are you sure you put the right files in the right GLPI (I mean if you have multiple GLPI instances) ?

ermannog:
no I have only one GLPI instances, 
but I solved in this way:
I disabled the plugin
I not delete the folder of the plugin but I overwrite the folder with the version 9.2 +2.0
Then manually delete these:
fusioninventory\lib\mustache.js-2.0.0
fusioninventory\pics\beta.png
fusioninventory\pics\Translate.png
fusioninventory\scripts\createSNMPWalks.php


I have GLPI 9.2.2 on IIS and another strange  things that I observed was that when I was try do:
- stop IIS
- delete the folder of plugin
- copy the folder of the new version fo the plugin
- start IIS
when I try connect to GLPI and error 500 was returned and the only way to recover was restore the entire glpi folder from a backup.

ddurieux:
And the error was?

ermannog:
From php-errors.log:

  From Plugin::getInfo() in .....\glpi\inc\plugin.class.php line 1374
  plugin_version_fusioninventory-9.2.1.0 method must be defined!

ddurieux:
9.2.1.0 ? so it's not version 9.2+2.0

ermannog:
that error was returned only when I replaced the directory with that of version 9.2. + 2.0

















Suggerisci una modifica












































































































































 


 

 

 


 

 

 

 


 

 

 

 


 
That error was returned only when I replaced the directory with that of version 9.2. + 2.0

[...]

PR #2554, opened 2018-05-28
Allow update action : Where condition required

camlafit:
Glpi DB instance require a where condition.

As it's applied to each row, a simple condition always true.

Related to #2553 and maybe #2544

### DIFF
--- a/install/update.php
+++ b/install/update.php
@@ -1610,7 +1610,9 @@ function do_entities_migration($migration) {
          $DB->update(
             'glpi_plugin_fusioninventory_entities', [
                'agent_base_url' => $agent_base_url
-            ], []
+            ], [
+               '1 = 1'
+            ]
          );
       }
    }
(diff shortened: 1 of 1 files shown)


install/update.php
```diff
@@ -1610,7 +1610,9 @@ function do_entities_migration($migration) {
          $DB->update(
             'glpi_plugin_fusioninventory_entities', [
                'agent_base_url' => $agent_base_url
-            ], []
+            ], [
+               '1 = 1'
```
ddurieux: prefer use `[1 => 1]` or `['1' => '1']` instead `'1 = 1'`

camlafit: Ok corrected

About other where condition as there is a parameter , I presume 1 = 1 clause should be wrong

camlafit:
Is invalid , see #2556

PR #2582, opened 2018-07-05
Fix to get collect fields in search engine when defined in parent ent…

ddurieux:
…ity and recusrsive

### DIFF
--- a/inc/collect.class.php
+++ b/inc/collect.class.php
@@ -164,7 +164,7 @@ class PluginFusioninventoryCollect extends CommonDBTM {
       $i = 5200;
 
       $pfCollect = new PluginFusioninventoryCollect();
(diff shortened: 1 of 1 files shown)


inc/collect.class.php
```diff
@@ -164,7 +164,7 @@ static function getSearchOptionsToAdd($itemtype = NULL) {
       $i = 5200;
 
       $pfCollect = new PluginFusioninventoryCollect();
-      foreach ($pfCollect->find(getEntitiesRestrictRequest("", $pfCollect->getTable()), 'id ASC') as $collect) {
+      foreach ($pfCollect->find(getEntitiesRestrictRequest("", $pfCollect->getTable(), '', '', true), 'id ASC') as $collect) {
```
trasher: Sounds good to me. BTW, you should not call `find` inside the loop ;)

ddurieux: I know, but here I only fix.
FOr the find, I think will do a PR only for all find in loop (I think there are many)

PR #2585, opened 2018-07-09
A few more iterator replacements

trasher:


inc/inventorycomputerlib.class.php
```diff
@@ -880,12 +1002,13 @@ function updateComputer($a_computerinventory, $computers_id, $no_history, $setdy
 
                if (isset($a_vm['uuid'])
                && $a_vm['uuid'] != '') {
-                  $query = "SELECT * FROM `glpi_computers`
-                        WHERE `uuid` ".ComputerVirtualMachine::getUUIDRestrictRequest($a_vm['uuid'])."
-                        LIMIT 1"; // TODO: Add entity search
-                  $result = $DB->query($query);
+                  $iterator = $DB->request([
+                     'FROM'   => 'glpi_computers',
+                     'WHERE'  => ['uuid' => ComputerVirtualMachine::getUUIDRestrictRequest($a_vm['uuid'])],
```
orthagh: getUUIDRestrictCriteria maybe ?

trasher: indeed, I changed to use same as core

PR #2604, opened 2018-07-23
Update de_DE.po

TW1920:
New German Translation added


trasher:
Use transifex for translations: https://www.transifex.com/ddurieux/FusionInventory/

PR #2566, opened 2018-06-12
Update search engine fields to be compatible with 9.3

ddurieux:

PR #2610, opened 2018-07-30
Fix computer search options

trasher:


trasher:
Following https://github.com/fusioninventory/fusioninventory-for-glpi/pull/2566/files#diff-96d83989ef62fbe6bcec408b4ed8b62aR69

PR #2624, opened 2018-08-13
Add Last boot translation and typo-correction Omport to Import

sbeneke:
Add translation for Last boot correct Typo-error Omport to Import


ddurieux:
You don;t need do this here, you must use transifex to translate : https://www.transifex.com/ddurieux/FusionInventory/dashboard/

PR #2641, opened 2018-09-06
Update phpunit

trasher:
Parent constructor must be called
Do not use phpunit 7...
Call ini_set at the beginning
Try to fix session init failure


trasher:
Tests are green; this sounds like a good start :D

Issue #2357, opened 2017-11-22
Problem in collecting the operating system


kleverson4g:
Good afternoon
I have the following problem
at the moment the agent brings information from the machine, does not come the operating system, instead of the operating system comes in the field software, however much I switch to the operating system I wish it back and would invent again for software.
Is there a possibility to correct this?
Thank you very much and I await.

Issue #2623, opened 2018-08-10
OS infos not updated after RHEL update


stonebuzz:
After update a server from RHEL 7.2 to RHEL 7.5
OS info are not updated.

No lock found on glpi_plugin_fusioninventory_locks.

Paradoxically, he updates the rpm, including the ones that indicate the kernel and the version of OS.

GLPI 9.2.2
FI 9.2+2.1

trasher:
See #2644

Issue #2596, opened 2018-07-16
Field operating system unchanged


mvestrotech:
When the operating system elements is changed manually and an invantary is started, the previously modified information remains unchanged.

Best Regard.

tabad:
Hello @mvestrotech,

That's because when you change a field by hand you lock it (it forbids the automatic update by FI4G). Whether you want the information will be automatically updated, don't change it by hand. Please, unlock the field you locked and force an inventory.

This is not an issue; that's the normal behavior.

Regards,

tabad:
Hello @mvestrotech,

You're welcome. By the way, could you close this issue?

Thanks in advance. Regards,

stonebuzz:
Hello @tabad , I have the same problem 

however I can not find an OS lock  in the tab "locks" of the PC or in the configuration of FI

How can i unlock OS field ?

GLPI 9.2.2
FI 9.2+2.0-RC1

Best regards

tabad:
Hello @stonebuzz,

You are right, there isn't the possibility of lock/unlock any field regarding operating system in GLPI v9.2.x. I'm, sorry, I thought it was like whatever other fields.

I have read your issue #2618 and I have remembered when I upgraded from GLPI 9.1.x to GLPI 9.2.x a lot of computers was linked with more than one operating system which I think is an upgrade error. Theoretically this query shouldn't show any registry but I had a lot of them.

```sql
select items_id, operatingsystems_id from glpi_items_operatingsystems where itemtype = 'Computer' group by items_id, operatingsystems_id having count(items_id) > 1;
```

I'll study the update of operating system information further. ;)

Regards,

PR #2644, opened 2018-09-07
use correct glpi rights to import xml

orthagh:
I'm not very confident with the impact of this.
But the rights injection in glpi session is clearly bad.

Fix: #2357, #2596, #2623


rschubnell:
We changed the communication.class.php as described.
But the Computer has still the wrong Operating System.
An there is a entry under Software: Microsoft Windows 10 Enterprise (45290)
Plugin_FusionInventory	Software	Install a software: Microsoft Windows 10 Enterprise - 10.0.15063 (299154)

trasher:
There is still on `$_SESSION['glpi_plugin_fusioninventory_profile']['unmanaged'] = 'w';` in `phpunit/Common_TestCase.php`

PR #2645, opened 2018-09-10
fix the repeated display of the switch name

brahimmachkouri:
example : it displayed SWITCH1$$$$SWITCH1 instead of SWITCH1. 
bug
After inserting DISTINCT in the query :
fixed


trasher:
Should not this one target 9.3 branch?

ddurieux:
Merged manually

PR #2650, opened 2018-09-12
Glpi9.3+phpunit for php7.2

ddurieux:


phpunit/1_Unit/ComputerUpdateTest.php
```diff
@@ -1105,6 +1105,9 @@ public function ComputerMonitor() {
           'is_recursive'      => '0'
       ];
 
+      echo gettype($a_reference);
```
trasher: I guess this is some debuging output, should be removed ;)

Issue #2653, opened 2018-09-13
Errors in the plugin using cli_install


DIOHz0r:
After using the cli_install script I have notices two main problems for the plugin

1. You can't see the plugin on the menu and if you try to enter it using the link on the plugin list you get a bad permission message.

2. If you install the plugin using a user the agent base url is empty.

PR #2654, opened 2018-09-13
Fixes for cli_install script

DIOHz0r:
This PR is to solve the installation problems reported in issue #2653


scripts/cli_install.php
```diff
@@ -104,9 +104,14 @@
    die("GLPI not installed\n");
 }
 
+$user = new User();
 if (!is_null($args['--as-user'])) {
-   $user = new User();
    $user->getFromDBbyName($args['--as-user']);
+} else if (is_null($args['--as-user'])) {
```
AnaelMobilia: Why not using *else { xxx }* instead ?

Regards,
Anael

DIOHz0r:
I need some help here to understand how the tests works because with the change made some of them are failing, I think the problem is with the hard coded url used on tests, could you review it and give me some advice to solve it?.

ddurieux:
We have a problem with the tests, they not pass since few days (we think a problem with GLPI code). I will check it in the week.

PR #2661, opened 2018-09-24
license_id has been renamed licenseid (not a FKEY field)

trasher:


trasher:
This fix is real and do pass the tests.
Remaining issues are once again specific to unit tests ; out of the scope of this pull request.

phpunit/bootstrap.php
```diff
@@ -5,7 +5,7 @@
 define('TU_USER', '_test_user');
 define('TU_PASS', 'PhpUnit_4');
 
-global $CFG_GLPI;
+global $CFG_GLPIi, $GLPI_CACHE;
```
orthagh: vim insertion mode :D

trasher: Good catch :1st_place_medal:

ddurieux:
Fixed in #2740

PR #2664, opened 2018-09-26
Fixes

trasher:


trasher:
Yes, FI tests are failing again on travis, but this is not related to the current pull request.

PR #2682, opened 2018-10-10
Update rule arguments for GLPI 9.3.2

DIOHz0r:
Hello,

This PR is for fixing a wrong method compatibility with GLPI 9.3.2 because a new argument added to the executeActions method.

imagen


DIOHz0r:
I found this PR is a duplicated of https://github.com/fusioninventory/fusioninventory-for-glpi/pull/2590 so that is why I close this.

PR #2687, opened 2018-10-15
[wait start 9.3+2.0] Update import rules to pass connections detected on switches (mac, CD…

ddurieux:
…P/LLDP).

So now, all pass in the rules, it's the 7 first rules that manage it (before the rule 'Computer constraint (name)')


ddurieux:
replaced by #2881

Issue #2698, opened 2018-10-26
Add computers_id to searchOptions of Rest API


sebeka:
Hello,

I have a Feature request.
Could you please add the computers_id in the searchOptions of class PluginFusioninventoryAgent?
That would be really nice!
With this feature, we would be able, to figure out the computer in database with the help of the agents device_id and the rest api  api.
We need this for adding some additional information requested by the human user. (Group of Computer, otherserial, ...)

Thanks very much!

Sebastian

sebeka:
Sorry, figured out now, that "PluginFusioninventoryAgent" is not unique and changes every time the agent is run.

PR #2704, opened 2018-11-01
Add computers ID to the searchOptions for rest API

sebeka:
This is a lite enhancement related to issue #2698 , adding the computers ID to the searchOptions of class PluginFusioninventoryAgent. This allows to find the computer by the help of the agent name when using the rest API.


sebeka:
sorry have to checkout 9.3 instead

PR #2721, opened 2018-11-27
Fix wrong version number in footer

doudz:


ddurieux:
I added a code (very few days ago) in our release script and check in this file, so thanks for the PR but it will be managed in the release process to not forget modify the version in the file.

PR #2740, opened 2019-01-07
Glpi9.4+fix tests

ddurieux:


inc/deploypackage.class.php
```diff
@@ -1646,14 +1646,14 @@ function getPackageForMe($users_id, $computers_id = false) {
       $pfDeployGroup = new PluginFusioninventoryDeployGroup();
       $my_packages   = []; //Store all installable packages
 
-      $query = "";
+      $query = [];
       if ($users_id) {
-         $query = "`users_id`='".$users_id."' AND ";
+         $query += ['users_id' => $users_id];
```
trasher: Just `$query['users_id'] = $users_id;` would be correct, no? (same for the following +=)

inc/deploytaskjob.class.php
```diff
@@ -274,9 +274,9 @@ static function getActions($params) {
                   break;
 
                case 'Group':
-                  $like = "1";
+                  $like = [];
                   if (isset($params['query'])) {
-                     $like = "name LIKE '%".$DB->escape($params['query'])."'";
+                     $like += ['name' => ['LIKE', '%'.$DB->escape($params['query'])]];
```
trasher: `$like['name'] = ...`

PR #2742, opened 2019-01-09
provide manufacturer name when importing computer's model in dictionnary

orthagh:
When a computer's model dictionary set a manufacturer criterion, it's never provided by fusioninventory.
Ex: 
image


ddurieux:
Seems good, perhaps add a little test ;)

orthagh:
Done

PR #2748, opened 2019-01-25
drives capacity can also be updated

orthagh:


orthagh:
ping @ddurieux .

Test fails on php 7.3 only but i think it's not due to this PR.

trasher:
Indeed, dependencies packages were not installed for the build :/ I've just re-reun the 7.3 build again, it gets stalled...

Well, this is not related to the current PR ;)

Issue #2785, opened 2019-03-07
netinventory-printers - xml vs what's displayed


Stoatwblr:
There are a few instances where the generated XML doesn't match what's being displayed in the "Fusioninventory SNMP" supplies status page - things are missing.

```
    <DEVICE>
      <CARTRIDGES>
        <FUSERKIT>48</FUSERKIT>
        <TONERBLACK>37</TONERBLACK>
        <TONERCYAN>61</TONERCYAN>
        <TONERMAGENTA>28</TONERMAGENTA>
        <TONERYELLOW>31</TONERYELLOW>
        <TRANSFERKIT>32</TRANSFERKIT>
      </CARTRIDGES>
      <FIRMWARES>
        <DATE>20131018</DATE>
        <DESCRIPTION>device firmware</DESCRIPTION>
        <MANUFACTURER>Hewlett-Packard</MANUFACTURER>
        <NAME>hp color LaserJet 5550</NAME>
        <TYPE>device</TYPE>
        <VERSION>07.220.1</VERSION>
      </FIRMWARES>
      <INFO>
```

For this one (and ones like it) the fuser and transferbelt kit status isn't showing.

```
   <DEVICE>
      <CARTRIDGES>
        <CARTRIDGECYAN>73</CARTRIDGECYAN>
        <CARTRIDGELIGHTCYAN>42</CARTRIDGELIGHTCYAN>
        <CARTRIDGELIGHTGRAY>70</CARTRIDGELIGHTGRAY>
        <CARTRIDGELIGHTMAGENTA>58</CARTRIDGELIGHTMAGENTA>
        <CARTRIDGEMAGENTA>43</CARTRIDGEMAGENTA>
        <CARTRIDGEMATTEBLACK>73</CARTRIDGEMATTEBLACK>
        <CARTRIDGEPHOTOBLACK>75</CARTRIDGEPHOTOBLACK>
        <CARTRIDGEYELLOW>54</CARTRIDGEYELLOW>
      </CARTRIDGES>
      <FIRMWARES>
        <DATE>20160629</DATE>
        <DESCRIPTION>device firmware</DESCRIPTION>
        <MANUFACTURER>Hewlett-Packard</MANUFACTURER>
        <NAME>HP Designjet Z5200 PostScript</NAME>
        <TYPE>device</TYPE>
        <VERSION>TR8-AC_17.0.0.1</VERSION>
      </FIRMWARES>
```

[...]

PR #2787, opened 2019-03-14
Printerinks

Stoatwblr:
This patch adds new mappings to enable display of the the following printer supply item statuses and adds new aliases for some mappings to cater to variations of wording 
 eg CARTRIDGE{COLOR}LIGHT vs CARTRIDGELIGHT{COLOR} and GREY vs GRAY

This is to address FI4G bug #2785 

I believe all the new mappings are new and unique. They work here with all tested printers.

-  All status items are already present in XML being sent from Fusioninventory-agent, but not parsed by FI4G

 - "FUSERKIT" (Emitted by many higher end HP laserjets and several other devices)
 - "TRANSFERKIT" output (Various laser printers)

- "CARTRIDGELIGHTMAGENTA" (variant on CARTRIDGEMAGENTALIGHT)
- "CARTRIDGELIGHTCYAN" (variant on CARTRIDGECYANLIGHT)
- "CARTRIDGEPHOTOBLACK" (variant on CARTRIDGEBLACKPHOTO)
- "CARTRIDGEGRAY" (variant on CARTRIDGEGREY)
- "CARTRIDGEGREYLIGHT" - and variants CARTRIDGEGRAYLIGHT CARTRIDGELIGHTGREY CARTRIDGELIGHTGRAY
- "CARTRIDGEBLACKMATTE" - and variant CARTRIDGEMATTEBLACK

- CARTRIDGELIGHTMAGENTA, CARTRIDGELIGHTCYAN, CARTRIDGELIGHTGRAY, CARTRIDGEPHOTOBLACK, CARTRIDGEBLACKMATTE and the normal CYAN/MAGENTA/YELLOW are all emitted by Fusioninventory-agent when scanning HP Designjet Z5200 printers - other colours may be emitted depending on which inks are loaded (there are multiple palettes and ink types), so adding those might be a good idea (or some way of autoadding ink colours as seen)

Added "CARTRIDGEGLOSSENHANCER", "CARTRIDGEBLUE", "CARTRIDGEGREEN", "CARTRIDGERED" and "CARTRIDGECHROMATICRED" to cover the remaining known type 70/772 pigment cartridges used in this printer and in HP 12-color models. Latex inks probably identify differently.


ddurieux:
Merged manually (and squashed to have only 1 commit).
I have renammed the id 197and 198 to 211 and 212 because 197 and 198 used by another PR merged before this.

PR #2801, opened 2019-03-28
Fix a security problem. Thanks to Julien Szlamowicz && Damien Picard …

ddurieux:
…from company synacktiv

### DIFF
--- a/inc/toolbox.class.php
+++ b/inc/toolbox.class.php
@@ -515,7 +515,8 @@ class PluginFusioninventoryToolbox {
     * @param string $itemtype
     */
    static function sendXML($items_id, $itemtype) {
-      if (call_user_func([$itemtype, 'canView'])) {
+      if (preg_match("/^([a-zA-Z]+)\/(\d+)\/(\d+)\.xml$/", $items_id)
+         && call_user_func([$itemtype, 'canView'])) {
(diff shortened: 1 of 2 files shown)


inc/toolbox.class.php
```diff
@@ -515,7 +515,8 @@ static function sendSerializedInventory($items_id, $itemtype) {
     * @param string $itemtype
     */
    static function sendXML($items_id, $itemtype) {
-      if (call_user_func([$itemtype, 'canView'])) {
+      if (preg_match("/^([a-zA-Z]+)\/(\d+)\/(\d+)\.xml$/", $items_id)
```
trasher: Cannot be sure because I'm not aware of the original issue; but what is the goal here? Just checking a file name/extension may not be enough.

ddurieux: for the extension it reduce but check too if not have `..` in the variable
The goal is to be sure we read the file in the right folder, this file is a name/an id/fileid.xml, it's always same format, that's why I check with regex.

The performance isn't really a problem because this function is used only when user download the XML from the asset form

trasher: OK, to fix a `..` issue, this is enough. Perfs was not a point ;)

PR #2841, opened 2019-07-04
error on ShowPackage function +CS pb settled

jdudon:


ddurieux:
You must add an indent (3 characters) in the lines between old `}` and the new `}`

PR #2865, opened 2019-09-18
Add an option to enable monitor serialnumber partial matching

g-bougard:
Option is disabled by default

As monitor serial numbers are very badly handled by manufacturers in EDID block, the one found by the agent is often partly the one visible on the device itself. When people manually adds monitors in its stock, it will of course put as serial number the one found on device itself or on its packaging.
So this option could be enabled when people starts to see duplicated monitors with serial numbers partially different. This won't merge duplicated entries, so people will always have to delete one of the duplicated entries (the one from the dynamic inventory as first choice).

### DIFF
--- a/inc/inventoryruleimport.class.php
+++ b/inc/inventoryruleimport.class.php
@@ -570,11 +570,11 @@ class PluginFusioninventoryInventoryRuleImport extends Rule {
                      OR `[typetable]`.`serial`='".$serial2."')";
                   $_SESSION["plugin_fusioninventory_serialHP"] = $serial2;
 
-               // Support partial match for monitor serial
                } else if (isset($input['itemtype'])
                        AND $input['itemtype'] == 'Monitor'
                        AND $pfConfig->getValue('import_monitor_on_partial_sn') == 1
                        AND strlen($input["serial"]) >= 4) {
+                  // Support partial match for monitor serial
                   $sql_where_temp = " AND `[typetable]`.`serial` LIKE '%".$input["serial"]."%'";
                } else {
                   $sql_where_temp = " AND `[typetable]`.`serial`='".$input["serial"]."'";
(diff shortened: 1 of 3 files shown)


inc/inventoryruleimport.class.php
```diff
@@ -568,6 +570,11 @@ function findWithGlobalCriteria($input) {
                      OR `[typetable]`.`serial`='".$serial2."')";
                   $_SESSION["plugin_fusioninventory_serialHP"] = $serial2;
 
+               // Support partial match for monitor serial
+               } else if (isset($input['itemtype'])
+                       AND $input['itemtype'] == 'Monitor'
+                       AND $pfConfig->getValue('import_monitor_on_partial_sn') == 1) {
+                  $sql_where_temp = " AND `[typetable]`.`serial` LIKE '%".$input["serial"]."%'";
```
trasher: I have no idea what a partial SN should look like, but isn't that a little too large?

g-bougard: Just added a condition on at least 4 chars on the serial lenght. This should be largely enough for the purpose and will avoid unwanted match on empty or corrupted serials.

trasher:
You have a cs issue, could you fix plase:
```
FILE: ...oninventory/glpi/plugins/fusioninventory/inc/inventoryruleimport.class.php

--------------------------------------------------------------------------------

FOUND 1 ERROR AFFECTING 1 LINE

--------------------------------------------------------------------------------

 573 | ERROR | [x] Line indented incorrectly; expected at least 18 spaces,

     |       |     found 15

--------------------------------------------------------------------------------

PHPCBF CAN FIX THE 1 MARKED SNIFF VIOLATIONS AUTOMATICALLY

--------------------------------------------------------------------------------
```

Issue #2863, opened 2019-09-16
Adapt for new software management in core


trasher:
See https://github.com/glpi-project/glpi/pull/6216

PR #2866, opened 2019-09-19
Reflect core changes

trasher:
#2863


inc/inventorycomputerlib.class.php
```diff
@@ -1093,9 +1093,9 @@ function updateComputer($a_computerinventory, $computers_id, $no_history, $setdy
                      }
 
                      // Encryption details
-                     $input['encryption_tool'] = $disk['encryption_tool'];
-                     $input['encryption_algorithm'] = $disk['encryption_algorithm'];
-                     $input['encryption_type'] = $disk['encrypt_type'];
+                     $input['encryption_tool'] = (isset($disk['encryption_tool']) ? $disk['encryption_tool'] : null);
```
orthagh:
```suggestion
                     $input['encryption_tool'] = $disk['encryption_tool'] ??  null;
```

same for the 2 following lines (null coalescing available in php7.0)

PR #2870, opened 2019-10-04
Fix problem with lock and subitem delete fileds. delete fields in inp…

ddurieux:
…ut are be done more later, so before add / update the input

### DIFF
--- a/phpunit/2_Integration/DevicesLocksTest.php
+++ b/phpunit/2_Integration/DevicesLocksTest.php
@@ -380,4 +380,133 @@ class DevicesLocks extends RestoreDatabase_TestCase {
    }
 
 
+   /**
+    * @test
+    */
+   public function testLockMonitor() {
+      global $DB;
+
+      $DB->connect();
+
+      $_SESSION['glpiactive_entity'] = 0;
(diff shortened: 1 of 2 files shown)


phpunit/2_Integration/DevicesLocksTest.php
```diff
@@ -380,4 +380,133 @@ public function switchLockItemtype() {
...
+              [
+                  'name'    => 'DELL P2010H',
+                  'manufacturers_id'=> $manufacturer->getID(),
+                  'serial'  => 'P125R03U243P',
+                  'is_dynamic' => 1
+              ]
+          ],
+      ];
+      $this->assertEquals($a_computerinventory, $reference, 'Replaceid with lock must not delete fields');
+   }
```
g-bougard: I think another test updating the computer should show the lock is working on the computer object.

ddurieux: you're right, I'm currently add it

ddurieux: Humm have yet it function computerManufacturerLock before this test

PR #2879, opened 2019-10-17
Fix netdiscovery/netinventory default timeout and threads number usage

g-bougard:
Actually timeout sent to the agent for netdiscovery & netinventory tasks only comes from the agent configuration. The "General Setup" is never used.

As the default value set in database when an agent is created is "0", and as a "0" timeout is generally meaning "use the default", I uses this value to decide to provide the "General setup" value in place to the agent.

On the agent side, "0" means "use default" and default is 60 seconds... This is not what is though in the plugin as the general setup default is "1" second for netdiscovery and "15" for netinventory.

This may lead to misunderstandings when a task seems to be stick but it is waiting a minute for each ip to respond...

In the PR, I uses `__('General setup')` as default text for "0", but I'm not sure this is the better to use.

### DIFF
--- a/inc/networkdiscovery.class.php
+++ b/inc/networkdiscovery.class.php
@@ -336,9 +336,16 @@ class PluginFusioninventoryNetworkdiscovery extends PluginFusioninventoryCommuni
       $sxml_param = $sxml_option->addChild('PARAM');
       $sxml_param->addAttribute('THREADS_DISCOVERY',
          $pfAgent->fields["threads_networkdiscovery"]);
-      $sxml_param->addAttribute('TIMEOUT',
-         $pfAgent->fields["timeout_networkdiscovery"]);
-      $sxml_param->addAttribute('PID', $jobstate->fields['id']);
+      // Use general config when timeout is set to 0 on the agent
+      if ($pfAgent->fields["timeout_networkdiscovery"]<1) {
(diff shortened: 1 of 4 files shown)


g-bougard:
Here a sample of the agent configuration panel thanks to this PR:
image

inc/networkdiscovery.class.php
```diff
@@ -327,18 +327,31 @@ function run($jobstate) {
...
       $sxml_option->addChild('NAME', 'NETDISCOVERY');
 
       $sxml_param = $sxml_option->addChild('PARAM');
-      $sxml_param->addAttribute('THREADS_DISCOVERY',
-         $pfAgent->fields["threads_networkdiscovery"]);
-      $sxml_param->addAttribute('TIMEOUT',
-         $pfAgent->fields["timeout_networkdiscovery"]);
-      $sxml_param->addAttribute('PID', $jobstate->fields['id']);
+      // Use general config when threads number is set to 0 on the agent
+      if ($pfAgent->fields["threads_networkdiscovery"]<1) {
```
ddurieux: Better to use `== 0` instead `< 1`

g-bougard: Okay, that should be clearer. I changed the 4 case where I used `<1` check.

ddurieux:
Merged manually

PR #2889, opened 2019-11-03
Fix travis execution error

ddurieux:


ddurieux:
Fix the problem with mysql

PR #2904, opened 2019-11-22
Revert "Optimize is_array"

trasher:
This reverts commit 68c6259061787bb68be9553e4b8d362c8f4d072b.

Since PHP 7, `is_array` is faster than casting
```
is_array  :  0.046628952026367
cast, === :  0.18426394462585

Tested 1000000 iterations.
```


ddurieux:
Same test with same number of iteration give me another result (but better with is_array) :
```
the value is a string:
is_array  :  0.020073890686035
cast, === :  0.044270992279053

the value is an array:
is_array  : 0.021353006362915
cast, === : 0.021230220794678
```

PR #2941, opened 2020-01-30
Fix problem on dynamic group use plugin fileds criteria, cron not get…

ddurieux:
… the criteria and so get all computers.

### DIFF
--- a/inc/computer.class.php
+++ b/inc/computer.class.php
@@ -72,7 +72,9 @@ class PluginFusioninventoryComputer extends Computer {
                'id'   => 'fields_plugin',
                'name' => __('Plugin fields')
             ];
-            $fieldsoptions =  plugin_fields_getAddSearchOptions('Computer');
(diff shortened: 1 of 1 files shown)


inc/computer.class.php
```diff
@@ -212,5 +214,118 @@ static function showMassiveActionsSubForm(MassiveAction $ma) {
...
+    * when cron run, the offical function return nothing because not have profile
+    */
+   function pluginFieldsGetAddSearchOptions($itemtype, $containers_id = false) {
+      global $DB;
+
+      $opt = [];
+
+      $i = 76665;
+
+      $query = "SELECT DISTINCT fields.id, fields.name, fields.label, fields.type, fields.is_readonly,
```
trasher: Using dbiterator would be much better.

ddurieux: Yes, the goal is not rewrite all the plugin field function, but copy it here and remove the 3 lines.
I will propose a complete patch in the plugin fields repository soon

PR #2934, opened 2020-01-24
Fix display type of group (search and in group form)

ddurieux:

PR #2942, opened 2020-01-30
show type dropdown in deploy group

orthagh:
since 991455459f5606a8b16b297a561a59d48af45420

the display false broke the deploy group form


ddurieux:
See #2934

Issue #2946, opened 2020-02-06
When we manually update the computer, fusion doesn't take into account the general locks


stonebuzz:
Use case : 

Before installing the agent on a workstation, it is pre-registered in GLPI in order to manage the beginning of the life cycle.

First step create a computer with : 

- model
- type
- manufacturer
- serial (for link rule)

![image](https://user-images.githubusercontent.com/7335054/73923968-ff7a7d80-48cb-11ea-8fbe-aed05d66db42.png)

Then, configure default status 

![image](https://user-images.githubusercontent.com/7335054/73924498-e2927a00-48cc-11ea-8bdd-586b14b24b76.png)

and configure general lock to prevent overload of manufacturer, model and type

![image](https://user-images.githubusercontent.com/7335054/73924343-9cd5b180-48cc-11ea-9422-d43d48daac1c.png)

Result in BDD

```
+----+----------------+----------+---------------------------------------------------------------------------+
| id | tablename      | items_id | tablefields                                                               |
+----+----------------+----------+---------------------------------------------------------------------------+
|  1 | glpi_computers |        0 | ["networks_id","computermodels_id","computertypes_id","manufacturers_id"] |
+----+----------------+----------+---------------------------------------------------------------------------+

```

Then, Install the agent on the workstation and run the inventory.

Fusion update computer with rule link by serial and we see general lock for manufacturer, model and type

![image](https://user-images.githubusercontent.com/7335054/73924676-29806f80-48cd-11ea-8ba7-6f9e0bd00339.png)

Fusion add two locks for this computer  on serial and name because it's defined before inventory (ok with it)

Result in DB

```
+----+----------------+----------+-------------------+
| id | tablename      | items_id | tablefields       |
+----+----------------+----------+-------------------+
| 25 | glpi_computers |        3 | ["serial","name"] |
+----+----------------+----------+-------------------+
```

Now change status

Fusion add lock for this field (nice)

```
+----+----------------+----------+-------------------------------+
| id | tablename      | items_id | tablefields                   |
+----+----------------+----------+-------------------------------+
| 25 | glpi_computers |        3 | ["states_id","serial","name"] |
+----+----------------+----------+-------------------------------+

```
But it clears or replace value by previous inventory values  for manufacturer / model and type because fusion doesn't take account of genreal lock.

In my case it clears because values not exist in previous inventory

[...]

PR #2947, opened 2020-02-06
Load general lock on manually item update

stonebuzz:
On manually item update load general lock to prevent clear or update values

see #2946 

Best regards

### DIFF
--- a/inc/lock.class.php
+++ b/inc/lock.class.php
@@ -734,10 +734,8 @@ class PluginFusioninventoryLock extends CommonDBTM{
       }
 
       // load general lock configuration
-      $db_lock = $DB->fetch_assoc(PluginFusioninventoryLock::getLock($item->fields['tablename'], 0));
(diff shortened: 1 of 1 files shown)


inc/lock.class.php
```diff
@@ -731,6 +733,12 @@ static function deleteLock($item) {
          }
       }
 
+      // load general lock configuration
+      $db_lock = $DB->fetch_assoc(PluginFusioninventoryLock::getLock($item->fields['tablename'], 0));
```
orthagh: this 3 lines can be replaced by 
```php
$lockItemtype = PluginFusioninventoryLock::getLockFields($item->fields['tablename'], 0);
```

Also, i think it could be preferable to call the var `$generalLocks`

stonebuzz: done

lolokai:
Any news abOut this bug ?

PR #2951, opened 2020-02-11
Duplicate with postshowtab hook usage

trasher:
Because I can see fusion infos in the Lock tab, but also on the Virtualization and the notes one


trasher:
Current fix also remove it on the lock tab

trasher:
I no longer have the issue oO Maybe cache issue switching two branchs

PR #2955, opened 2020-02-17
Enhance wake agent when an agent not available and defined on many tasks

ddurieux:


ddurieux:
Merged into 10.0.2 version

PR #2958, opened 2020-02-19
Fix#2286: default states_id should not erase locked states_id

realitix:
Backport commit efd5eb0377629281328903fbe14e19917b923d7e into 9.1 branch


trasher:
9.1 is no longer maintained

PR #2962, opened 2020-02-26
Try with db transaction instead of db restore

trasher:
Just a try.


trasher:
Too many errors to get that working

Issue #3003, opened 2020-07-17
/inc/printerlog.class.php   .getDateRequest


Cycu85:
hello 

getDateRequest() was deleted from GLPI 9.5

but is still use in inc/printerlog.class.php at line 557

Issue #3002, opened 2020-07-16
massive action "target a task" from computer form doesn't work


FraFau:
![image](https://user-images.githubusercontent.com/59955425/87684105-234ac500-c782-11ea-92aa-a388488a4b3f.png)

the result is : 
![image](https://user-images.githubusercontent.com/59955425/87684289-5f7e2580-c782-11ea-858b-f8be6ab71e80.png)

and in php-errors.log : 
`[2020-07-16 16:35:15] glpiphplog.CRITICAL:   *** Uncaught Exception TypeError: Argument 1 passed to Dropdown::addNewCondition() must be of the type array, string given, called in /var/www/glpi/inc/dropdown.class.php on line 157 in /var/www/glpi/inc/dropdown.class.php at line 278
  Backtrace :
  inc/dropdown.class.php:157                         Dropdown::addNewCondition()
  plugins/fusioninventory/inc/task.class.php:1925    Dropdown::show()
  inc/massiveaction.class.php:661                    PluginFusioninventoryTask::showMassiveActionsSubForm()
  ajax/dropdownMassiveAction.php:48                  MassiveAction->showSubForm()
`

FraFau:
Thank you !

PR #3004, opened 2020-07-17
Fixes

trasher:
See #3002 and #3003


trasher:
Proposed fix for #3003 is a quick hack; real solution is to rework queries; but nothing has really been done on fusion for that; and I'm not going to do it myself.

Issue #3013, opened 2020-08-12
[9.5.0+1.0] Execution of taskscheduler is blocked when fields plugin (1.11.0) is installed through Marketplace


TuX1sPuNk:
### Informations
In GLPI, taskscheduler task status is blocked on **Running**.
FusionInventory For GLPI version : **9.5.0 + 1.0**
Fields version : **1.11.0**

### Issue
After manualy launching **cron.php**
![Capture](https://user-images.githubusercontent.com/69008422/89997356-53578c00-dc8c-11ea-9886-d12c07db3591.PNG)
_The FatalError is caused by having the Fields plugin installed in plugins folder and marketplace folder (after upgrading through marketplace) (needs to be ignored for issue as I had corrected the problem for next test)_
The task status is again blocked on **Running**
In **computer.class.php**, the error is in the foreach that add additional fields to search but on line 70, the include look for : `GLPI_ROOT . "/plugins/fields/hook.php"`, as with GLPI 9.5+, plugins can be installed in **marketplace** and therefore installed in the **marketplace** folder not in **plugins** folder.
Also even if Fields plugin is only installed in plugins folder, we still have the warning  : invalid arguments for the foreach
`PHP Warning: Invalid argument supplied fo foreach() in /var/www/html/plugins/fusioninventory/inc/computer.class.php on line 76`

### Test made
Removing Fields plugin folder from the plugins folder, and having it installed in marketplace folder
![Capture2](https://user-images.githubusercontent.com/69008422/89999066-9dda0800-dc8e-11ea-8616-14701ad4ee68.PNG)
**Not working**, tries to look for the plugin in plugins folder only.

Removing Fields plugin folder from marketplace folder, and having it installed in plugins folder
![Capture3](https://user-images.githubusercontent.com/69008422/89999493-2c4e8980-dc8f-11ea-9e6e-ffdd2257beb4.PNG)
**Task status isn't blocked on Running**, but still have the warning.

Deactivating Fields plugin
**Task is working with no errors**

### Solution
Add a check if the plugin Fields is installed through marketplace or manually in plugins folder and include from the correct path according to the check

ddurieux:
Not use the marketplace...

TuX1sPuNk:
Not using the marketplace is just a temporary fix.
I have made a pull request with a fix to this issue so people can use the marketplace or manually install plugin as they want

PR #3014, opened 2020-08-12
Fix for issue #3013 : Added check if Fields plugin is installed through the marketplace

TuX1sPuNk:
Added check to test if Fields plugin is installed through the marketplace, so installed in marketplace folder not plugins folder.
If so include hook.php from marketplace path instead of plugins path.

#3013

### DIFF
--- a/inc/computer.class.php
+++ b/inc/computer.class.php
@@ -67,7 +67,12 @@ class PluginFusioninventoryComputer extends Computer {
       $plugin = new Plugin();
       if ($plugin->isInstalled('fields')) {
          if ($plugin->isActivated('fields')) {
-            include_once(GLPI_ROOT . "/plugins/fields/hook.php");
+            // Check if Fields plugin is installed in marketplace or plugins folder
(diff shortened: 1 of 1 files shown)


inc/computer.class.php
```diff
@@ -67,7 +67,12 @@ function rawSearchOptions() {
       $plugin = new Plugin();
       if ($plugin->isInstalled('fields')) {
          if ($plugin->isActivated('fields')) {
-            include_once(GLPI_ROOT . "/plugins/fields/hook.php");
+            // Check if Fields plugin is installed in marketplace or plugins folder
```
trasher, line 70: The `Plugin::getPhpDir()` method could be used here.

TuX1sPuNk: Didn't knew this function, I've made the change

trasher: No problem ;) It's been introduced in 9.5 especially for that ;)

trasher (approved):

orthagh (approved):

PR #3024, opened 2020-08-28
Fix typo equipment

koshikawa:
Fix typo equipment

### DIFF
--- a/inc/config.class.php
+++ b/inc/config.class.php
@@ -466,7 +466,7 @@ class PluginFusioninventoryConfig extends CommonDBTM {
       echo "</tr>";
 
       echo "<tr class='tab_bg_1'>";
-      echo "<td>".__('Network equipment')."</td>";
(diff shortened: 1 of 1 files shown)


inc/config.class.php
```diff
@@ -466,7 +466,7 @@ function changestatus() {
       echo "</tr>";
 
       echo "<tr class='tab_bg_1'>";
-      echo "<td>".__('Network equpment')."</td>";
+      echo "<td>".__('Network equipment')."</td>";
```
trasher, line 469: This string does not seems to exists in core, so:
```suggestion
      echo "<td>".__('Network equipment', 'fusioninventory')."</td>";
```

trasher (changes requested):
See comment

koshikawa:
> See comment

OK

trasher (approved):
Thank you

orthagh (approved):

PR #3042, opened 2020-10-08
Fix battery handling

btry:
see 

incomplete spec  for JSON format
https://github.com/glpi-project/inventory_format/pull/18

XML spec 
https://github.com/fusioninventory/fusioninventory.github.io/pull/75

Bad value collected by the agent
https://github.com/fusioninventory/fusioninventory-agent/issues/840
https://github.com/fusioninventory/fusioninventory-agent/pull/843

support for current max capacity to monitor battery wear
https://github.com/glpi-project/glpi/pull/8062

### DIFF
--- a/inc/inventorycomputerlib.class.php
+++ b/inc/inventorycomputerlib.class.php
@@ -2287,6 +2287,9 @@ class PluginFusioninventoryInventoryComputerLib extends PluginFusioninventoryInv
       $data['capacity']           = $data['capacity'];
       $data['voltage']            = $data['voltage'];
       $batteries_id = $deviceBattery->import($data);
+      if (!isset($data['real_capacity'])) {
+         $data['real_capacity'] = '0';
+      }
       $data['devicebatteries_id'] = $batteries_id;
       $data['itemtype']           = 'Computer';
       $data['items_id']           = $computers_id;
--- a/inc/inventorycomputerlib.class.php
+++ b/inc/inventorycomputerlib.class.php
@@ -1362,7 +1362,7 @@ class PluginFusioninventoryInventoryComputerLib extends PluginFusioninventoryInv
                      }
 
                      if ($update === true) {
-                        $item_DeviceMemory->update($input);
+                        $item_DeviceBattery->update($input);
                      }
 
                      unset($a_computerinventory['batteries'][$key]);
--- a/inc/inventorycomputerlib.class.php
+++ b/inc/inventorycomputerlib.class.php
@@ -1342,28 +1342,11 @@ class PluginFusioninventoryInventoryComputerLib extends PluginFusioninventoryInv
                      && isset($arraydb['serial'])
                      && $arrayslower['serial'] == $arraydb['serial']
                   ) {
-                     $update = false;
-                     if ($arraydb['capacity'] == 0
-                              && $arrayslower['capacity'] > 0) {
-                        $input = [
-                           'id'       => $keydb,
-                           'capacity' => $arrayslower['capacity']
-                        ];
-                        $update = true;
-                     }
-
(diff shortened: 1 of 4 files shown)


inc/inventorycomputerlib.class.php
```diff
@@ -1342,28 +1342,11 @@ function updateComputer($a_computerinventory, $computers_id, $no_history, $setdy
...
-                     }
-
-                     if ($arraydb['voltage'] == 0
-                              && $arrayslower['voltage'] > 0) {
-                        $input = [
-                           'id'        => $keydb,
-                           'voltage'   => $arrayslower['voltage']
-                        ];
-                        $update = true;
-                     }
```
btry, line 1362: Voltage belongs to DeviceBattery, not Item_DeviceBattery. It is useless to handle it here as the column does not exists.

inc/inventorycomputerlib.class.php
```diff
@@ -1342,28 +1342,11 @@ function updateComputer($a_computerinventory, $computers_id, $no_history, $setdy
...
-
-                     if ($update === true) {
-                        $item_DeviceMemory->update($input);
-                     }
+                     $input = [
+                        'id' => $keydb,
+                        'real_capacity' => $arrayslower['real_capacity'],
+                     ];
```
btry, line 1348: Support for real_capacity, which beslongs to Item_DeviceBttery

inc/inventorycomputerlib.class.php
```diff
@@ -1342,28 +1342,11 @@ function updateComputer($a_computerinventory, $computers_id, $no_history, $setdy
...
-                              && $arrayslower['voltage'] > 0) {
-                        $input = [
-                           'id'        => $keydb,
-                           'voltage'   => $arrayslower['voltage']
-                        ];
-                        $update = true;
-                     }
-
-                     if ($update === true) {
-                        $item_DeviceMemory->update($input);
```
btry, line 1365: bad object here

trasher:
Please fix tests

trasher (approved):
Sounds good to me, even if not tested (waiting for next agent release)

cedric-anne (approved):

btry:
@trasher : To test the latest agent you can clone the repo and run 

```
perl bin/fusioninventory-agent -s https://server/glpi/plugins/fusioninventory
```

or 
```
perl bin/fusioninventory-agent -l /tmp/inventory.xml
````
then import the xml in Fusion Inventory.

**Comment by trasher** [comment]
Code assumes capacity (and voltage) is an integer, but in my case, this is not:
```
    <BATTERIES>
      <CAPACITY>29.07 Wh</CAPACITY>
      <CHEMISTRY>lithium-polymer</CHEMISTRY>
      <DATE>10/11/2015</DATE>
      <MANUFACTURER>SMP</MANUFACTURER>
      <NAME>DELL JHXPY53</NAME>
      <SERIAL>3525</SERIAL>
      <VOLTAGE>8.497 V</VOLTAGE>
    </BATTERIES>
```

Result in glpi:
image

trasher:
(using fusion agent 2.5.2)

g-bougard:
@trasher If you're still using agent 2.5.2, this XML part is simply wrong. So not the plugin fault, but the agent one. Using agent 2.6 will fix this problem.

trasher:
Agent data is not wrong, but not the correct unit

g-bougard:
Indeed the agent did not respect the defined specs and the specs tell to provide an integer normalized to mWh for CAPACITY and mV for VOLTAGE. So sorry, but in your case, the agent was really wrong as it should have normalize values and only send the raw values without unit.

trasher:
All is fine then

g-bougard (approved):

PR #3054, opened 2020-10-23
Add Force Inventory in massiveaction

OseiasMiller:


ddurieux:
We never wanted to integrate this feature because it's can be a problem with server performances.
If you wakeup 900 computers, you will have 900 agent connections on the serveur and it can be a very bad idea.
For this reason, we will not integrate this feature / PR, sorry.

PR #3066, opened 2020-12-09
feat(core): add queued inventory process

stonebuzz:
Add capability to manage inventories in  queued

### DIFF
--- a/front/queuedinventory.form.php
+++ b/front/queuedinventory.form.php
@@ -1,0 +1,78 @@
+<?php
+
+/**
+ * FusionInventory
+ *
+ * Copyright (C) 2010-2016 by the FusionInventory Development Team.
+ *
+ * http://www.fusioninventory.org/
+ * https://github.com/fusioninventory/fusioninventory-for-glpi
+ * http://forge.fusioninventory.org/
+ *
+ * ------------------------------------------------------------------------
+ *
+ * LICENSE
+ *
+ * This file is part of FusionInventory project.
+ *
+ * FusionInventory is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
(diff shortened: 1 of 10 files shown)


ddurieux (changes requested):
What is the goal of this?

stonebuzz:
The final goal is to have two modes of import: direct and indirect.

The indirect mode, allows to store XML in database (it will be processed later by an auto task and clean).

XML integration can be handled during the day rather than when the agent contacts the plugin.

This avoids server overloads when all the agents come up at the same time.

ddurieux:
Not sure manage them later (queue) is a solution if you have all inventories in same time.
Why do you have inventories in same time, with 24h agent prolog running, at 8h-10h you can have about 20 - 30% of inventories in same time.

In case you have few hours for agent prolog, you will have queue fill and never empty, so you move the problem.

You can't replace server ressources and optimization of webserver && database per a queue, and in context, it will introduce another problems instead fix the initial problem.

I'm not agree to introduce this queue...

stonebuzz:
It's only an option, I don't force anyone to use it.

g-bougard:
@ddurieux What is proposed here is to have an optional way to answer a real problem. You know the 24h prolog is still a problem the monday morning when people comes back to work on very large infrastructures.
I think this option can help to by-pass load peaks.

g-bougard:
Personally I would prefer to see such queue system on disk like Postfix does for mail queueing or at least to have the choice between fs & db.

ddurieux:
And perhaps it could be managed outside the plugin (I mean before), like the fusioninventory collect script => http://fusioninventory.org/documentation/fi4g/dmz/

ddurieux:
I close for the reasons I have written

PR #3111, opened 2021-03-04
Fix "Import" and "Cleanup unused files" deploy packages link

eduardomozart:
Fix "Import" and "Cleanup unused files" deploy packages link. Fix issue #3069.


ddurieux:
Not works in all cases, fix here: #3117

PR #3110, opened 2021-03-04
Fix broken paths

AdrienClairembault:
Issue introduced by #3107, a lot of paths are broken due to a variable being overridden: 

image

PR #3113, opened 2021-03-05
Fix $fi_path override

romarioliveira25:
That fix solve a trouble found with the override of the variable $fi_path at the line changed at this propused change


ddurieux:
Fixed in #3110

PR #3152, opened 2021-04-16
Fill domain criterion before applying location rules

### LINKED ISSUE #3151: Domain criterion gets empty in Location Rules
Recently we found that the Domain criterion in the location rules is always empty so we cannot use it at all.

We wanted to use the Location Rules to automatically assign a location to new and existent computers on each inventory agent run. The location of our computers is encoded in computer hostnames. So we created a new location rule, selected `Domain` as a criterion and specified an action to set a particular location. But that rule just didn't work, we tried different conditions but haven't succeeded. After many tries, it gets obvious that the criterion is just empty so the location rules cannot work.

Then we found that entity rules also have the `Domain` criterion and an action to set location and luckily it worked as expected so we can set/update location on agent runs.

The question is why is the `Domain` criterion empty when it is used in the location rules?

Version:
GLPI 9.5.4 Copyright (C) 2015-2021 Teclib' and contributors
FusionInventory 9.5+3.0 - Copyleft © 2010-2019 by FusionInventory Team

vykulakov:
Close #3151

This change fills the `domain` field of the input data before passing it to the location rule processor. It allows users using the `Domain` criterion in location rules.

### DIFF
--- a/inc/inventorycomputerinventory.class.php
+++ b/inc/inventorycomputerinventory.class.php
@@ -327,7 +327,12 @@ class PluginFusioninventoryInventoryComputerInventory {
 
       //Add the location if needed (play rule locations engine)
       $output = [];
-      $output = PluginFusioninventoryToolbox::addLocation($input, $output);
+      $inputloc = $input;
+      if ((isset($a_computerinventory['Computer']['domains_id']))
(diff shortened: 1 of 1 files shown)


ddurieux (approved):

ddurieux:
Thanks for the fix ;)

PR #3172, opened 2021-05-25
fix(formatconverter): use LIKE keyword to match user

stonebuzz:
Match user with LIKE keyword
in case when glpi login user is
'testfusion@xxxx.com'
but FusionInventory receive from XML 

```
<DOMAIN>YYYYY</DOMAIN>
<LOGIN>tesfusion</LOGIN>
```

Internal : 22094


ddurieux:
Your path will have false positive.
If I understand, you want search on `testfusion@xxxx.com` and `testfusion` ?

stonebuzz:
nop, i want to search fusion user (```testfusion```) from glpi users format (```testfusion@xxxx.com```)

In case when fusion domain is not the same as GLPI

ddurieux:
so `['LIKE' , $array_tmp['login'].'@%']]` no? (with @).
The problem is if you have in users `toto`, `tototi` , `toto@foo.bar`, it will find `tototi` with your code and it's not the right user.
You see the problem ?

stonebuzz:
yes  I see

PR #3230, opened 2021-09-27
Fix collect task reporting on not found file collect

g-bougard:
By now, when a file collect task finds nothing, it just reports the task run "Successful" and there's no way to understand the file search indeed failed.
With this update, the task goes in error when no file is found and we have a better reporting when more than one search file job is defined:
image

### DIFF
--- a/inc/collect.class.php
+++ b/inc/collect.class.php
@@ -786,13 +786,18 @@ class PluginFusioninventoryCollect extends CommonDBTM {
                      break;
 
                   case 'file':
+                     if ($sid) {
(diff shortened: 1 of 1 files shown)


g-bougard:
In the case of file search, we should probably remove the "Info" line as it finally duplicates the "Running" one.

g-bougard:
Hi @ddurieux 
it's been 6 months I opened this PR. Do you think it could be merged at some time in the future ?
Otherwise can you just close it ?
Thanks

g-bougard:
Closing without maintainer answer and as he published a new release. I can only guess he doesn't mind.

PR #3292, opened 2022-03-10
A version to zip file for fusioninventory

jhonatas-mendes:
The tar command on the DIOXX container image doesn't work properly and that's why I created a zip decompression version


ddurieux:
What is this???

jhonatas-mendes:
Hello Duriexx, I'm not able to unzip any plugin in the container at the address "/var/www/html/glpi/plugins" so the suggestion to add this type of zip as well.  And the news is that I didn't get it that way either 😂

ddurieux:
Ah ok, so you need an archive in zip format, that's it?

Issue #3316, opened 2022-04-07
Wrong monitor name and model


mva-klp:
Hi!
I have a problem with import and update monitors.
In monitor's history:
377879 | 06-04-2022 16:32 | Plugin_FusionInventory (6) | Модель | Изменение с 2476WM (48) на **\&nbsp; (2476W)**
377878 | 06-04-2022 16:32 | Plugin_FusionInventory (6) | Наименование | Изменение с AOC L2476 VWM на **24/2016**
![image](https://user-images.githubusercontent.com/84769092/162160932-d0f20d81-b637-4928-8c61-12711d91c889.png)
![image](https://user-images.githubusercontent.com/84769092/162161092-20a42e68-379d-4efb-abe6-d81684cde45b.png)

I use GLPI 9.5.7
FusionInventory 9.5+4.0
FusionInventory agent for Windows x64 2.6

ddurieux:
it's the name get in the inventory.
The model has been added in model form

mva-klp:
I get this \&nbsp; (2476W) and field model in result is empty
![image](https://user-images.githubusercontent.com/84769092/162164485-a280ae6e-0be9-43fc-b71f-a81e2d4593e5.png)
 and how can i will get normal monitor's name instead of **24/2016**

ddurieux:
not normal :(
Can you give me part of <MONITORS> section of the XML file?

mva-klp:
for example one of computers XML:
`<MONITORS>
      <BASE64>AP///////wBaYyTYAQEBAQgVAQMYNB14Lu7VpVVImyYSUFS/74CzAKlAlQ+VAIGAgUBxTwEBAjqAGHE4LUBYLEUACSUhAAAeAAAA/wBTM1UxMTA4MDA5ODMKAAAA/QAySxhSEQAKICAgICAgAAAA/ABWQTI0MzEgU2VyaWVzAJ0=</BASE64>
      <CAPTION>VA2431 Series</CAPTION>
      <DESCRIPTION>8/2011</DESCRIPTION>
      <MANUFACTURER>ViewSonic Corporation</MANUFACTURER>
      <PORT>VGA</PORT>
      <SERIAL>S3U110800983</SERIAL>
    </MONITORS>
    <MONITORS>
      <ALTSERIAL>1950HLNC500474</ALTSERIAL>
[...]

PR #3317, opened 2022-04-08
Combining of two fields for name's field of monitor

mva-klp:
I've added combining of two fields: the first word from the string in MANUFACTURER + CAPTION. I think that it is good idea. And this decides the first problem in issue https://github.com/fusioninventory/fusioninventory-for-glpi/issues/3316


ddurieux (approved):

ddurieux:
Some tests failed, can you check them? 
or I can do it in couple days if you don't have time / don't know how to fix them / other reasons :D

PR #3356, opened 2022-06-29
getTaskjobstatesForAgent  order by list of targets

SebastienBatteur:
the result should be ordered by the future job.index field when drag AND drop feature will be properly activated in the taskjobs list.


ddurieux:
Thanks for the patch, sorry for the too long delay to merge it.

